PR #6, opened 2020-07-18
Use *const c_char instead of *const i8

luxxxxy:
This will fail to compile on targets such as `riscv64gc-unknown-linux-gnu` otherwise.

### DIFF
--- a/src/ffi.rs
+++ b/src/ffi.rs
@@ -1,6 +1,6 @@
 #![allow(non_camel_case_types)]
 
-use std::os::raw::c_void;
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Only x86 is supported, but ok.

luxxxxy:
... I got resvg/tiny-skia/rustybuzz to successfully compile and run in a RISC-V emulator, but ok.

RazrFalcon:
By support, I mean that I can guarantee that it works correctly.

PR #16, opened 2020-11-09
Fix memory leak.

laurmaedje:
The function `rb_face_get_table_data` referenced table data, which creates a new shallow subblob. This blob was never destroyed though. The function is used in `crate::ot::get_table_data` and that's in turn used to get the table data in all places where we just receive a `rb_face_t` without a `ttfp_face`. Since the blob is just a shallow subblob it's pointer stays valid even when we destroy it. This is obviously a bit of a hack, but the whole `get_table_data` is a workaround anyway.


RazrFalcon:
Yes, I was also confused with `hb_blob_t`.

Travis is very slow today...

PR #17, opened 2020-11-11
Merge font and face

laurmaedje:
This PR merges `rb_font_t`, `rb_face_t`, `rb_ot_face_t`, `Font` and `Face` into a single pair of `Face` / `rb_face_t` that is stored entirely on the Rust side. 

The seven tables that are still parsed in C++ are sanitized only once upon `Face` creation, but not lazily as in the C++ version. I implemented the lazy version first, but it turns out that most of the tables are sanitized anyways (e.g. `kerx` when the `rb_ot_map_t` is constructed due to a call to `rb_aat_layout_has_positioning`), so I think that complicates things unnecessarily.

I also removed a lot of unused code, some of it not directly related to the font/face change. I hope you don't mind that; if it makes reviewing harder, please let me know.


RazrFalcon:
Nice!

PR #19, opened 2020-11-15
Small refactoring

laurmaedje:
Here are a a few small improvements. I removed the unused buffer functions, used `GlyphInfo::{as_char, as_glyph}` for less `try_from + unwrap`, wrapped the AAT and kern functions I missed before and moved the horizontal-to-vertical char mapping from `shape.rs` to `unicode.rs`.

PS: Would you want an `aat` feature-gate upstream or not? Otherwise, I would just remove AAT from my fork. And do you plan to make a new crates.io release soon or only once AAT is done?


src/unicode.rs
```diff
@@ -700,6 +701,57 @@ impl CharExt for char {
         (0x0FE00..=0x0FE0F).contains(&ch) || // VARIATION SELECTOR - 1..16
         (0xE0100..=0xE01EF).contains(&ch)    // VARIATION SELECTOR - 17..256
     }
+
+    fn vertical(self) -> Option<char> {
+        Some(match u32::from(self) >> 8 {
+            0x20 => match self {
+                '\u{2013}' => '\u{fe32}', // EN DASH
```
RazrFalcon, line 708: Would you mind keeping hex coded? Unicode notation is too ugly/verbose.

RazrFalcon:
I think we can make a release with the `aat` feature. Remember to make the `cc` dependency optional and make a dummy `main` in the `build.rs`

RazrFalcon:
I've just run the benchmarks, and Arabic shaping become 7x slower. I don't really care about the performance right now, but maybe there is an easy fix? I guess this is because we're not using caching and lazy loading.

src/unicode.rs
laurmaedje: Sure.

laurmaedje:
The kerning functions would then also be behind the `aat` feature, right?

With the slowdown, I think there are mostly two factors at play:
1. Harfbuzz's parsing approach is fundamentally faster because it sanitizes once and then just accesses the tables unchecked. The way I wrote GSUB/GPOS and the way all of `ttf_parser` is written of course creates a lot more control flow. There are maybe some small improvements one could make here (i.e. checking stream length early so that multiple stream reads only need one check), but overall that's hard to fix.
2. That I removed the lookup acceleration with coverage sets. I don't know for sure, but I would assume that this would help a lot. I don't think it's _that_ much work to implement, depending on whether we would use harfbuzz's bit-optimized set or some data structure from Rust's standard library or from crates.io.

src/unicode.rs
laurmaedje: Only on the left side of the match though, right?

RazrFalcon: Ufff... you're returning char. Ok, leave it as is.

RazrFalcon:
Does we still have kerning in C++? I guess it would be a good idea to hide AAT test behind the `aat` feature too. But since tests are autogenerated it could be a problem. I think you can mark them manually for now and I will figure out what to do with them later.

As for performance, the current parsing approach should not be a bottleneck. `ttf_parser` is actually pretty fast, despite all the checks. Previously, we weren't that slow. I think you should try using coverage sets first and I will try profiling rustybuzz.

RazrFalcon:
Looks like the most time is spent in `bool OT::GSUBGPOS::sanitize<OT::PosLookup>(rb_sanitize_context_t*) const`. And the heaviest and most often called method is `bool OT::OffsetTo<OT::Anchor, OT::IntType<unsigned short, 2u>, true>::sanitize<>(rb_sanitize_context_t*, void const*) const`. It's called 10149 times on `benches/arabic.txt`.

src/unicode.rs
laurmaedje: Well, the method is only used with `u32` at the call site. But I thought conceptually it would fit well into `CharExt`.

laurmaedje:
Keeping parsed data just like that without allocations is not really possible. Because GSUB/GPOS basically consists of offset-lists to offset-lists etc. 

Specifically, GSUB contains
- a LookupList, which contains a variable number of offsets to
- Lookups, which contain a variable number of offsets to
- Lookup subtables and these subtables often have again one or two layers of variable nesting

We could replace all `Offsets16`s by `Vec`s. That would probably help performance overall. I didn't to that in the first place, because I wanted to try the `ttf_parser` style zero-allocation approach. 

It might also be the case that the parsing isn't the real bottleneck, that `ChainContextLookup::parse` is just called 10502 times while the shaper tries to apply the lookup over and over again, while it doesn't actually do anything most of the time. The coverage sets could really help here I think.

Another thing to note of which I'm not sure how it affects performance: [Matching returns a struct containing a big stack-allocated array](https://github.com/RazrFalcon/rustybuzz/blob/39aed5c8c1db826275043dd9705a5f0686e1da5d/src/ot/matching.rs#L46). Using a smallvec or taking the output array as a `&mut` argument might be better.

src/ot/kern.rs
```diff
@@ -8,6 +8,18 @@ use super::TableIndex;
...
+    unsafe { ffi::rb_ot_layout_has_machine_kerning(face.as_ptr()) != 0 }
+}
+
+pub fn has_cross_kerning(face: &Face) -> bool {
+    unsafe { ffi::rb_ot_layout_has_cross_kerning(face.as_ptr()) != 0 }
+}
+
 pub fn kern(plan: &ShapePlan, face: &Face, buffer: &mut Buffer) {
     unsafe { ffi::rb_ot_layout_kern(plan.as_ptr(), face.as_ptr(), buffer.as_ptr()); }
 }
```
laurmaedje, line 25: These four methods are still [implemented in C++](https://github.com/RazrFalcon/rustybuzz/blob/master/harfbuzz/src/hb-kern.cc) because they depend on the `kern` table (which in turn depends on `kerx`) and the `rb_aat_apply_context_t`.

src/unicode.rs
RazrFalcon: Yes, we should prefer char to u32. It's fine.

src/ot/kern.rs
RazrFalcon: What is the question?

RazrFalcon:
Yes, I know that GSUB/GPOS cannot be done in a zero-allocation way, because we're running through the same data again and again. Therefore using `Vec` to store `Lookup::subtables` should be a good start.

src/ot/kern.rs
laurmaedje: No question. That was an answer to your question whether we still have kerning in C++.

RazrFalcon: Oh, I see. So we cannot have a pure Rust version even with AAT feature-gated?

laurmaedje: Only if we put kerning behind the `aat` feature _or_ put it behind a `kern` feature that depends on the `aat` feature _or_ port it over from ttf-parser.

RazrFalcon: Hmm... In this case I suggest using your own fork for now, and I will try to port AAT as soon as possible.

laurmaedje: Okay, then I'll do that. I will also have a look at making GSUB/GPOS faster.

Is there anything still blocking this PR from getting merged?

RazrFalcon: Nope. Thanks. I will start working on AAT next week. Just to make sure we would not overlap.

PR #20, opened 2020-12-04
Improve OpenType Layout performance

laurmaedje:
I finally got around to work on the bad OpenType performance and I was able to improve performance quite a lot with relatively little changes! 

The first change was to keep all GSUB/GPOS lookups and their subtables in vectors instead of re-parsing them all the time. This improved performance across the board, especially the arabic shaping benchmark by a factor of 4.5x. 

The second change was to collect the coverage of all subtables in a lookup into a glyph set upfront. This set is checked before any complex apply/would_apply machinery runs. I did not port harfbuzz's bit-based set. Instead, I first experimented with a simple sorted vector of glyphs and Rust's hash set, but performance was not so great, so in the end I wrote a simple set based on a sorted `Vec` of ranges and binary search. The effect of this change is a bit less clear cut. It gives a good boost for the arabic shaping benchmark (around 1.6x faster), but makes latin shaping slower. But Latin shaping is, interestingly, still a lot faster than through `harfbuzz-rs`. I wonder what's the reason for that.

### Benchmarks
**Harfbuzz (through harfbuzz-rs)**
```
test shape_arabic_hb ... bench:     786,963 ns/iter (+/- 66,127)
test shape_latin_hb  ... bench:     136,060 ns/iter (+/- 9,707)
test shape_zalgo_hb  ... bench:     196,338 ns/iter (+/- 16,224)
```

**Rustybuzz before any of my PRs** (`5244b48`)
```
test shape_arabic_rb ... bench:     793,146 ns/iter (+/- 33,016)
test shape_latin_rb  ... bench:      76,425 ns/iter (+/- 4,093)
test shape_zalgo_rb  ... bench:     218,218 ns/iter (+/- 29,286)
```

**Rustybuzz master** (`043a0b5`)
```
test shape_arabic_rb ... bench:   6,302,550 ns/iter (+/- 104,301)
test shape_latin_rb  ... bench:      58,398 ns/iter (+/- 1,932)
test shape_zalgo_rb  ... bench:     517,315 ns/iter (+/- 14,301)
```

**Improvement: Parsing lookups and subtables only once and storing them in vectors** (`6332c81`)
```
test shape_arabic_rb ... bench:   1,426,730 ns/iter (+/- 13,970)
test shape_latin_rb  ... bench:      37,537 ns/iter (+/- 1,091)
test shape_zalgo_rb  ... bench:     270,590 ns/iter (+/- 24,512)
```

**Improvements: Parsing only once + collecting combined coverage of a lookup's subtables into a glyph set** (`f3f1a16`)
```
test shape_arabic_rb ... bench:     875,941 ns/iter (+/- 11,062)
test shape_latin_rb  ... bench:      66,402 ns/iter (+/- 1,957)
test shape_zalgo_rb  ... bench:     274,015 ns/iter (+/- 5,692)
```

### DIFF
--- a/src/complex/indic.rs
+++ b/src/complex/indic.rs
@@ -373,6 +373,7 @@ impl IndicWouldSubstituteFeature {
             let lookup = map.lookup(TableIndex::GSUB, index);
             let ctx = WouldApplyContext { glyphs, zero_context: self.zero_context };
             if face.gsub
+                .as_ref()
                 .and_then(|table| table.get_lookup(lookup.index))
                 .map_or(false, |lookup| lookup.would_apply(&ctx))
             {
--- a/src/face.rs
+++ b/src/face.rs
@@ -308,14 +308,14 @@ impl<'a> Face<'a> {
         }
     }
 
-    pub(crate) fn layout_table(&self, table_index: TableIndex) -> Option<SubstPosTable<'a>> {
+    pub(crate) fn layout_table(&self, table_index: TableIndex) -> Option<&SubstPosTable<'a>> {
         match table_index {
-            TableIndex::GSUB => self.gsub.map(|table| table.0),
-            TableIndex::GPOS => self.gpos.map(|table| table.0),
+            TableIndex::GSUB => self.gsub.as_ref().map(|table| &table.inner),
+            TableIndex::GPOS => self.gpos.as_ref().map(|table| &table.inner),
         }
     }
 
-    pub(crate) fn layout_tables(&self) -> impl Iterator<Item = (TableIndex, SubstPosTable<'a>)> + '_ {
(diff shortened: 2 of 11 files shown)


RazrFalcon:
Thanks! I've planed to work on this this weekends, but looks like I don't have to anymore.
And I've almost finished AAT. I have to port just one morx subtable and we're ready for a release.

RazrFalcon:
As for the Latin performance, note that we have a very different TrueType parsing logic. Maybe in this case it makes a difference. Hard to tell.

laurmaedje:
> And I've almost finished AAT. I have to port just one morx subtable and we're ready for a release.

 I'm looking forward to it :)

Issue #23, opened 2020-12-05
API Qs/feedback


dhardy:
Hey @RazrFalcon, just looking over your API for use in KAS-text and I have a few comments.

I notice that `Face` is a wrapper around `ttf-parser::Face`, which KAS-text already has an instance of. I can supply the data pointer and face index no problem, but I could also directly supply a `&'a Face` (or possibly even just store the `rustybuzz::Face` the embedded `ttf-parser::Face` could still be accessed). But this isn't really important.

More significant, KAS-text stores text-runs with a `FontId` (index in internal list) plus a size (`DPU` aka pixels per font unit). I see your `Face` object embeds the size. For my purposes would storing a single `Face` and adjusting the size before each shaping run be the best option?

Your `GlyphBuffer` reports `codepoint` and `cluster`. KAS-text needs the text-index of each glyph (required for editing support), and trying to reconstruct this from codepoints is not ideal.

I see you use the same `UnicodeBuffer ←→ GlyphBuffer` model as HarfBuzz. I'm not sure how best to cache this type since it is consumed by value in the `shape` method. Maybe it would be better if `UnicodeBuffer` and `GlyphBuffer` were wrappers around `Box<Buffer>` to avoid large copies? Or does the optimiser avoid this issue anyway?

RazrFalcon:
Sounds good. I will add `from_face` method.

Whoops, I forgot to mention in the readme that unlike hb, rb doesn't support font size. You should scale the produced offsets and advances manually.

>Your GlyphBuffer reports codepoint and cluster.

Same as harfbuzz. There are no changes here.

>Or does the optimiser avoid this issue anyway?

`UnicodeBuffer` and `GlyphBuffer` are just safe wrappers around internal `Buffer`. There is not copying involved. You can store/cache any of them.

dhardy:
> You should scale the produced offsets and advances manually.

So it's not necessary to set font size in the `Face` object at all, if not using variation axes?

> Same as harfbuzz.

Ach ja. Those names are confusing IMO. From KAS-text's HarfBuzz support:
```rust
        let index = idx_offset + info.cluster;
        assert!(info.codepoint <= u16::MAX as u32, "failed to map glyph id");
        let id = GlyphId(info.codepoint as u16);
```

The `Buffer` object is rather large. Put the below in godbolt.org (sorry, share feature isn't working!), and its clear there is a copy (compare different sizes of `T`):
```rust
type T = [u32; 32];
[...]

PR #34, opened 2021-04-06
Creating Face from ttf_parser::Face

lain-dono:
Mostly solves #23. via implementing `Font::from_face` and `AsRef`/`AsMut` to `ttf_parser::Face` and `Deref`/`DerefMut` to `ttf_parser::FaceTables`.

All this allows use of `Face` as a drop-in replacement for `ttf_parser::Face`.

In addition, I updated the dependencies and made `Face::units_per_em` public.

### DIFF
--- a/src/face.rs
+++ b/src/face.rs
@@ -57,18 +57,18 @@ impl<'a> AsMut<ttf_parser::Face<'a>> for Face<'a> {
 }
 
 impl<'a> core::ops::Deref for Face<'a> {
-    type Target = ttf_parser::FaceTables<'a>;
+    type Target = ttf_parser::Face<'a>;
 
     #[inline]
     fn deref(&self) -> &Self::Target {
-        core::ops::Deref::deref(&self.ttfp_face)
+        &self.ttfp_face
     }
 }
 
 impl<'a> core::ops::DerefMut for Face<'a> {
     #[inline]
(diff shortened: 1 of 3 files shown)


src/face.rs
```diff
@@ -41,34 +42,87 @@ pub struct Face<'a> {
...
+}
+
+impl<'a> AsMut<ttf_parser::Face<'a>> for Face<'a> {
+    #[inline]
+    fn as_mut(&mut self) -> &mut ttf_parser::Face<'a> {
+        &mut self.ttfp_face
+    }
+}
+
+impl<'a> core::ops::Deref for Face<'a> {
```
RazrFalcon, line 59: Shouldn't it return `ttf_parser::Face`?

RazrFalcon (changes requested):

src/face.rs
lain-dono: No. This duplicates the functionality of `ttf_parser::Face`. This allows the ascender/descender/etc to be retrieved directly.

RazrFalcon: A user should work with `ttf_parser::Face`, not `ttf_parser::FaceTables`. Or you simply doesn't want a double-deref?

lain-dono: Hmm, I thought it wouldn't generate documentation, but it turns out it does!

RazrFalcon:
Thanks!

PR #49, opened 2022-04-20
Fix `no_std` support by using `libm` for rounding

CryZe:
Turns out that `no_std` wasn't properly supported as `f32::round` isn't available on targets that don't have std. As with other crates, there's now two mutually exclusive features `std` and `libm` that you have to choose.

Additionally this applies a bunch of clippy fixes.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,11 +16,21 @@ exclude = ["benches/**"]
 bitflags = "1.2"
 bytemuck = { version = "1.5", features = ["extern_crate_alloc"] }
 smallvec = "1.6"
-ttf-parser = "0.15"
+ttf-parser = { version = "0.15", default-features = false, features = [
+    "opentype-layout",
+    "apple-layout",
+    "variable-fonts",
+    "glyph-names",
+] }
 unicode-bidi-mirroring = "0.1"
 unicode-ccc = "0.1.2"
 unicode-general-category = "0.4"
 unicode-script = "0.5.2"
+libm = { version = "0.2.2", optional = true }
+
+[features]
+default = ["std"]
+std = []
 
 [dev-dependencies]
 pico-args = "0.3"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -5,6 +5,12 @@ A complete [harfbuzz](https://github.com/harfbuzz/harfbuzz) shaping algorithm po
 #![no_std]
 #![warn(missing_docs)]
 
+#[cfg(not(any(feature = "std", feature = "libm")))]
+compile_error!("You have to activate either the `std` or the `libm` feature.");
+
+#[cfg(feature = "std")]
+extern crate std;
+
 extern crate alloc;
 
 #[macro_use]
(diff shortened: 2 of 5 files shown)


src/common.rs
```diff
@@ -561,7 +561,7 @@ impl core::str::FromStr for Feature {
 
             // Parse postfix.
             let had_equal = p.consume_byte(b'=').is_some();
-            let value1 = p.consume_i32().or(p.consume_bool().map(|b| b as i32));
+            let value1 = p.consume_i32().or_else(|| p.consume_bool().map(|b| b as i32));
```
CryZe, line 564: This is interesting as before you could write `kern=1off` and it would parse. That probably was a bug.

src/lib.rs
```diff
@@ -5,34 +5,52 @@ A complete [harfbuzz](https://github.com/harfbuzz/harfbuzz) shaping algorithm po
...
-pub use crate::buffer::{
-    GlyphPosition, GlyphInfo, BufferClusterLevel,
-    SerializeFlags, UnicodeBuffer, GlyphBuffer
+pub use crate::{
+    buffer::{
+        BufferClusterLevel, GlyphBuffer, GlyphInfo, GlyphPosition, SerializeFlags, UnicodeBuffer,
+    },
+    common::{script, Direction, Feature, Language, Script, Variation},
+    face::Face,
+    shape::shape,
```
CryZe, line 42: Oops, I still had rustfmt on here. I can revert this.

src/lib.rs
```diff
@@ -5,34 +5,52 @@ A complete [harfbuzz](https://github.com/harfbuzz/harfbuzz) shaping algorithm po
...
+fn round(x: f32) -> f32 {
+    #[cfg(feature = "std")]
+    {
+        x.round()
+    }
+    #[cfg(not(feature = "std"))]
+    {
+        libm::roundf(x)
+    }
+}
```
CryZe, line 56: I can move this into its own module.

RazrFalcon:
Thanks! But could you please remove clippy fixes? Or move them into a separate patch.

CryZe:
Alright I removed the clippy fixes.

Cargo.toml
```diff
@@ -16,11 +16,21 @@ exclude = ["benches/**"]
 bitflags = "1.2"
 bytemuck = { version = "1.5", features = ["extern_crate_alloc"] }
 smallvec = "1.6"
-ttf-parser = "0.15"
+ttf-parser = { version = "0.15", default-features = false, features = [
```
RazrFalcon, line 19: I don't think this change is necessary.

RazrFalcon (changes requested):

Cargo.toml
CryZe: `std` is enabled by default on ttf-parser 0.15:
```toml
default = ["std", "opentype-layout", "apple-layout", "variable-fonts", "glyph-names"]
```

RazrFalcon (approved):

CryZe:
Thank you so much :D

PR #57, opened 2022-10-23
Don't publish test data to crates.io

Shnatsel:
Brings down the crate size from 1.2MB to 230KB.

Right now anyone running `cargo install something-depending-on-rustybuzz` will have to also download rustybuzz's test data, which is useless to them and just wastes bandwidth.

If anyone intends to develop rustybuzz, they'll clone the git repo instead of fetching the crate from crates.io (which doesn't even have a UI for doing that).


RazrFalcon:
Makes sense.

Issue #44, opened 2021-11-17
Is it possible to push `char` into `UnicodeBuffer`?


Riey:
I want push some chars into UnicodeBuffer for now, I have to collect into String then use `push_str`

however, `push_str` seems just decode it into chars so I think `fn push(ch: char)` method can be possible or at least `fn push_chars(chars: impl Iterator<Item = char>)`

RazrFalcon:
Probably yes. Will have to take a look. In general, you have to preprocess your string beforehand.

bluebear94:
I’m interacting with Java code, so a way to bypass conversion into UTF-8 would be helpful for me, too.

PR #61, opened 2022-11-12
Add UnicodeBuffer::add method for adding a codepoint directly

bluebear94:
This addresses #44.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1569,6 +1569,13 @@ impl UnicodeBuffer {
         self.0.push_str(str);
     }
 
+    /// Appends a character to a buffer with the given cluster value.
+    #[inline]
+    pub fn add(&mut self, codepoint: char, cluster: u32) {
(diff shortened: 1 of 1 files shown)


src/buffer.rs
```diff
@@ -1569,6 +1569,12 @@ impl UnicodeBuffer {
         self.0.push_str(str);
     }
 
+    /// Appends a character to a buffer with the given cluster value.
+    #[inline]
+    pub fn add(&mut self, codepoint: char, cluster: u32) {
+        self.0.add(codepoint as u32, cluster)
```
RazrFalcon, line 1575: You have to call 

```
self.0.context_len[1] = 0;
```

as well.

https://github.com/harfbuzz/harfbuzz/blob/630f09c8b6022184f7a414bb5e07c01898fae60c/src/hb-buffer.cc#L1355

RazrFalcon:
Thanks.

PR #73, opened 2023-07-22
Exchange reverse_range() for the stdlib reverse()

notgull:
The lowest possible hanging fruit.


RazrFalcon:
Thanks! 
This project has a weird balance between having an idiomatic Rust code and looking as close as possible to the original implementation for ease of backporting changes.

Issue #37, opened 2021-06-27
Sync with harfbuzz 2.7.4


RazrFalcon:
Complex changes: 

- [ ] https://github.com/harfbuzz/harfbuzz/commit/6a38adeb577444cd164eeb6ace70741fe1a33791

PR #75, opened 2023-08-17
Sync with HarfBuzz 2.7.4

bluebear94:
See #37.

* [X] Update scripts
  * [X] `gen-tag-table.py`
  * [x] `gen-universal-table.py`
  * [ ] Refactor scripts (extract common logic, store downloaded data files in a separate directory) – maybe another PR?
* [X] Update remaining code
  * [X] Revise `universal_machine::find_syllables` to avoid the `Vec` allocation?
* [X] Update tests


bluebear94:
2.7.4’s `gen-universal-table.py` doesn’t run on the current version of the Unicode data files. (`gen-tag-table.py` also needed a change to run properly.) Would it be viable to update these to the latest version right away?

RazrFalcon:
@bluebear94 Good progress. Here is my failed attempt. Maybe it would help a bit: [Patch.patch](https://github.com/RazrFalcon/rustybuzz/files/12368928/Patch.patch)

I haven't been able to grok `machine_index_t` iterator logic. Afaik it's mutating and bidirectional, which I'm not sure how to implement in Rust.

Either way, feel free to ask me anything. Just remember that I wrote this code 3 years ago and I do not remember much myself.

RazrFalcon:
There is also a pseudo-bug. [This code](https://github.com/RazrFalcon/rustybuzz/blob/master/src/complex/universal_machine.rs#L370-L373) code should use `use_category`, not `indic_category`. It doesn't matter right now. But harfbuzz switched to `use_category` "byte" from 2 to 3.

RazrFalcon:
The way I was doing the back-porting is by checking each commit, not the whole diff at once. harfbuzz commits are usually pretty small and self-contains, so it's easier that way.

If a commit has changes related to build system, fonts parsing, subsetting, C++ stuff - it can be safely ignored.

If a commit contains changes to tests, like [this](https://github.com/harfbuzz/harfbuzz/commit/420a74f7725bec2d3d4688fda226163b06f9a564) one, then you have to checkout this commit. Build harfbuzz from sources:

```sh
meson builddir --reconfigure
ninja -Cbuilddir
```

 And then run `scripts/gen-shaping-tests.py /path/to/harfbuzz-src`

bluebear94:
HarfBuzz 2.7.4’s buffer serialization was [changed to surround the contents in square brackets](https://github.com/harfbuzz/harfbuzz/commit/c0716bb5dc8b25ec452e71afc74ebebbc927fed8). Should this be reflected in `rustybuzz`? Currently, it is, but this will require a minor change to `gen-shaping-tests.py`.

RazrFalcon:
>Should this be reflected in rustybuzz?

Sure.

bluebear94:
Trying to fix the remaining test failure. For the font `tests/fonts/in-house/8228d035fcd65d62ec9728fb34f42c63be93a5d3.ttf`, HarfBuzz is getting −639 for the *x*-bearing of glyph \#4’s extents, but rustybuzz is getting −638 instead.

scripts/gen-shaping-tests.py
```diff
@@ -99,7 +99,7 @@ def convert_test(hb_dir, hb_shape_exe, tests_name, file_name, idx, data, fonts):
     glyphs_expected = subprocess.run(options_list, check=True, stdout=subprocess.PIPE)\
         .stdout.decode()
 
-    glyphs_expected = glyphs_expected[1:-2]  # remove `[..]\n`
+    glyphs_expected = glyphs_expected[:-1]  # remove trailing newline
```
RazrFalcon, line 102: I think this one should stay as before. No need for needless `[]`

RazrFalcon:
I've checked that font with [ttf-explorer](https://github.com/RazrFalcon/ttf-explorer) and it does use `-639`. 
Is this a new test case or some recent change broke it?

scripts/gen-shaping-tests.py
bluebear94: This is the minor change to `gen-shaping-tests.py` I alluded to. :)

RazrFalcon: It overwrites all the rust tests, which is better to avoid.

bluebear94:
This test was supposed to have an expected output of `x=0+1030|acutecomb=0@-19,-27+0|X=2+1295|acutecomb=2@-151,320+0` (as generated by the `gen-shaping-tests.py` script), but the one currently in the master branch has `x=0+1030|acutecomb=0@-20,-27+0|X=2+1295|acutecomb=2@-152,321+0` instead. This test might have been manually changed after generation.

RazrFalcon:
Ugh... good find. I would assume this change was caused by [this commit](https://github.com/RazrFalcon/ttf-parser/commit/cbd8198504413634ab72ae0aea9784c634645c0d). In which case it's not a bug a we should prefer rustybuzz output. Some floats rounding differences are expected.

Lets leave the manual correction in place and I will think what could be done here.

bluebear94:
Now `morx_34_001` and `morx_36_001` are failing (producing shorter outputs than expected). These test resource exhaustion cases, and the failures might be related to the [work limit increases](https://github.com/harfbuzz/harfbuzz/commit/1c05f6789b86ae7bf54fb74255b3c8cea31d9978) in HarfBuzz.

RazrFalcon:
Rounding errors are ok, a different output isn't. 

How do you generate Rust tests? As I've said before, the ideal workflow is to checkout each harfbuzz commit you're trying to port, rebuild harfbuzz and rerun `scripts/gen-shaping-tests.py`. harfbuzz tests are not set in stone, they change and you have to constantly keep the Rust one up to date.

bluebear94:
I’ve identified and fixed the cause of `morx_34_001` and `morx_36_001` failing (HB increased buffer resource limits).

I’ve generated the Rust tests as you’ve specified (except that this time, I went directly to the 2.7.4 release). I think porting the changes one commit at a time might be more helpful in the future, though, so that when something goes wrong with the tests, I have less stuff to search through.

RazrFalcon:
Yes, harfbuzz tests are very tricky and pedantic, which is a good thing, but this also makes porting more tedious. 
You do have to port stuff commit-per-commit. Trying to port in bulk would rarely work in my experience. 
And when porting commit-per-commit you immediately know the cause.

When writing this crate originally, I was running tests after porting each C++ function. Otherwise you would never find the cause.

bluebear94:
Could you review my initial attempt to get rid of the `Vec` allocation in `find_syllables` at d523ea2?

I’m not completely satisfied with the use of `Cell`; I think I can avoid it by passing a reference to the buffer explicitly to `included`, `next_glyph`, and `prev_glyph`.

RazrFalcon:
Yes, I was struggling with this one as well. My initial implementation was almost identical to your iterator based one. But tests were failing.

Will take a look what can be done here.

src/complex/universal_machine.rs
```diff
@@ -422,3 +391,34 @@ fn found_syllable(
         *syllable_serial = 1;
     }
 }
+
+fn included(i: usize, buffer: &Buffer) -> bool {
```
RazrFalcon, line 395: Can we pass just the `&[GlyphInfo]` to those functions?

src/complex/universal_machine.rs
```diff
@@ -422,3 +391,34 @@ fn found_syllable(
...
+                return !glyph2.is_unicode_mark();
+            }
+        }
+    }
+    true
+}
+
+fn next_glyph(p: usize, buffer: &Buffer) -> usize {
+    let mut q = p + 1;
+    while q < buffer.info.len() && !included(q, buffer) {
```
RazrFalcon, line 412: Isn't this basically [Iterator::position](https://doc.rust-lang.org/stable/std/iter/trait.Iterator.html#method.position). Maybe it would be shorter and more clear. 

And `rposition` below.

Just a suggestion.

bluebear94: I’ve changed these to use `Iterator::find`.

src/complex/universal_machine.rs
```diff
@@ -392,33 +393,27 @@ fn found_syllable(
...
 }
 
-fn prev_glyph(p: usize, buffer: &Buffer) -> usize {
-    let mut q = p - 1;
-    while !included(q, buffer) {
-        q -= 1;
-    }
-    q
+fn prev_glyph(p: usize, infos: &[GlyphInfo]) -> usize {
+    (0..p).rev().find(|q| included(*q, infos)).unwrap()
```
RazrFalcon, line 418: I would prefer `unwrap_or(0)` here. Just in case.

bluebear94:
Merged the changes from #77 into this branch. Unfortunately, due to limitations in Ragel and in Rust, I had to switch back to allocating a vector in `universal_machine::find_syllables` for now.

RazrFalcon:
@bluebear94 Sure, no problem. Thanks for looking into it. Hopefully it was relatively easy for you to figure out how to build master ragel. I will update the readme soon.

RazrFalcon:
PS: I wasn't notified that this pull requires is no longer a draft. I would suggest pinging me directly. 
Do you consider it to be done, feature-wise?

bluebear94:
Yes.

I’ve almost figured out how to get rid of the vec allocation again, but unfortunately, it requires manually applying a few edits to the generated file. The problem is that Ragel outputs assignment statements such as `p = 0;`, but Rust doesn’t support overloading assignment unlike C++.

RazrFalcon:
Hm... I'm not familiar with Ragel myself, but it would be great to be able to avoid manual edits. Once again, there is no rush.

RazrFalcon:
By the way, why there are so many changes to `tests/shaping_text_rendering_tests.rs`?

bluebear94:
These are due to changes in the `morx_34_001` and `morx_36_001` tests due to increased resource limits in HarfBuzz 2.7.4 (mentioned in [this comment](https://github.com/RazrFalcon/rustybuzz/pull/75#issuecomment-1684455505)). None of the other tests in that file have changed.

RazrFalcon:
Hm... github shows:

>23,060 additions, 5 deletions not shown because the diff is too large. Please use a local Git client to view these changes.

I can't even see what was changed.

RazrFalcon:
>Refactor scripts (extract common logic, store downloaded data files in a separate directory) – maybe another PR?

Are you talking about Python scripts? I think that should stay untouched. They are ported from harfbuzz as well. Just reformatted and outputting Rust instead of C++. rustybuzz should have as little changes as possible.

behdad:
> > Refactor scripts (extract common logic, store downloaded data files in a separate directory) – maybe another PR?
> 
> Are you talking about Python scripts? I think that should stay untouched. They are ported from harfbuzz as well. Just reformatted and outputting Rust instead of C++. rustybuzz should have as little changes as possible.

Newer HarfBuzz versions use PackTab for a bunch of stuff. It's a small one-file Python library. Should be easy to add Rust output to it:

https://github.com/harfbuzz/packtab/issues/5

scripts/gen-tag-table.py
```diff
@@ -689,6 +698,9 @@ def get_name(self, lt):
 bcp_47.names['mhv'] = 'Arakanese'
 bcp_47.scopes['mhv'] = '(retired code)'
 
+ot.remove_language_ot('MONT')
+ot.add_language('mnw', 'MONT')
+
```
notgull, line 703: This doesn't appear in the original set of Harfbuzz changes, what is this for?

scripts/gen-tag-table.py
```diff
@@ -664,6 +673,8 @@ def get_name(self, lt):
 ot.remove_language_ot('IRT')
 ot.add_language('ga-Latg', 'IRT')
 
+ot.add_language('hy-arevmda', 'HYE')
+
 ot.remove_language_ot('KGE')
 ot.add_language('und-Geok', 'KGE')
```
notgull, line 680: This block of code:

```py
ot.add_language('guk', 'GUK')
ot.names['GUK'] = 'Gumuz(SIL fonts)'
ot.ranks['GUK'] = ot.ranks['GMZ'] + 1
```

should be removed

notgull (commented):
Took a look at the first couple commits. I'll work through the rest over the next week or so.

scripts/gen-tag-table.py
bluebear94: The current version of [Microsoft’s page of language tags](https://docs.microsoft.com/en-us/typography/opentype/spec/languagetags) has a note for Thailand Mon that chokes the parser. I had to add this for the script to finish successfully.

notgull (approved):
Sorry that it took me so long. From my remaining read-through it looks reasonably accurate to Harfbuzz 2.7.4.

In the future, for ease of review I think we should have a better system for porting over commits:

- Port over commits from Harfbuzz one by one.
- Link the commit being ported in the `rustybuzz` commit.

RazrFalcon:
@bluebear94 Any updates on why `tests/shaping_text_rendering_tests.rs` diff is so big?

@notgull Porting commits one by one is fine. Adding the exact commit hash is probably unnecessary. We simply don't need this information.

I plan to try to refactor the code base a bit to make it more harfbuzz-like. File structure, code structure, etc. Ideally, rustybuzz should look like harfbuzz transpiled to Rust. The less changes we have, the either is should be to port.

bluebear94:
> Any updates on why `tests/shaping_text_rendering_tests.rs` diff is so big?

See [my previous comment on this](https://github.com/RazrFalcon/rustybuzz/pull/75#issuecomment-1693203883). It might be possible to make the offending tests shorter using [`str::repeat`](https://doc.rust-lang.org/nightly/std/primitive.str.html#method.repeat), but that would require nontrivial changes to `gen-shaping-tests.py`. Alternatively, we can just exclude the `morx_34_001` and `morx_36_001` tests from being generated.

RazrFalcon:
@bluebear94 Oh, I think I got it now. Those tests produce an enormous output? And harfbuzz tests simply use `*` to indicate _all glyphs_? Is this right?

In this case I would add those tests to the ignore list in `scripts/gen-shaping-tests.py` for now and later I would try to figure out what to do with them.

bluebear94:
It’s not that all glyphs from these fonts are being tested, but rather than these fonts contain glyphs that expand indefinitely.

Anyway, I pushed a commit to ignore these tests.

RazrFalcon:
I see now. Those tests were already absurd before. I guess I would have to handle them separately. 

Anyway, thanks for your work! I honestly wasn't expecting anyone to be brave enough to do it.
I plan to do some small changes in the next couple of days and make a new release and then, if you're still interested, you can continue working on it. I just wanted to warn you not to fork it until I restructure the code base a bit.

RazrFalcon:
@bluebear94 If you're still interested in updating rustybuzz, I can give you write access to this repo, to simplify your workflow. This would allow more gradual changes.
Just make sure each commit passes the CI.

Issue #74, opened 2023-08-12
Deprecation


RazrFalcon:
@laurmaedje Hi! I'm giving up on this project. I have no further plans on working on it. I know you're using it in `typst`, so it's probably affects you.
I plan to either archive it or pass to someone else.

As for `resvg`, which is the reason this projects exists, I haven't decided yet. But basically I have a "choice" of using this deprecated version for now, switch to harfbuzz bindings or try out `swash` (I'm very skeptical).

Keavon:
Sad to see this :(

I know [COSMIC Text](https://github.com/pop-os/cosmic-text) is also built upon rustbuzz so it might be worth pinging @jackpot51 as well.

I noticed you called out [Swash](https://github.com/dfrg/swash) but didn't mention [Allsorts](https://github.com/yeslogic/allsorts). The latter is still maintained. I'm not very familiar with either of them in terms of their maturity and features, but I was wondering if there's a reason you didn't mention Allsorts or comment on its status by comparison to Swash.

notgull:
I would be willing to take over maintenance for bug fixes (although I probably won't add any features or make any significant optimizations). As `cosmic-text` and `theo` also both depend on this crate, I would like to see this maintained, and I'm sure there are many others who would as well.

@RazrFalcon It seems this crate is in the same boat as some of your other crates, like `tiny-skia`: used extensively throughout the ecosystem yet you no longer have the time/energy to maintain it. How would you feel about creating a GH organization, transferring some of your repos there, and then adding some interested volunteers to maintain those crates? I would definitely be happy to participate given the option.

Keavon:
This news might also be worth posting to Reddit. It's an invaluable part of the ecosystem and that could garner contributors. But I definitely see how...

> Since v2.7.1, harfbuzz received 5813 commits. 

...could be a pretty daunting proposition to keep up with.

RazrFalcon:
@Keavon cosmic-text also supports swash, so they should be fine. Both shapers are basically dead anyway.

I do mention `allsorts` at the end of the readme. It should be pretty good, but it lacks some features (variable fonts for one) and focuses on subsetting.

In general, it's extremely hard to compare various shaping libraries. I guess running one agains harfbuzz's test suite is the only option right now. Because it's the golden standard.

>This news might also be worth posting to Reddit.

@notgull  I need someone who would be back-porting harfbuzz changes. Which no one probably will. Otherwise there is no point.

[...]

PR #76, opened 2023-08-19
Replace hand-translated machines with nom

notgull:
The `_machine.rs` files are unmaintainable, as they are a 1:1 handwritten mirror of the machine-generated state machine files for Harfbuzz. My goal is to make this crate more maintainable by writing these files in such a way that it corresponds with the `.rl` files. This way, changes to the `.rl` files can be easily reflected in `rustybuzz`.

My original attempt to replace the `indic_machine.rs` file, however, has not passed tests. I am not familiar with `ragel` itself and I am mystified by its semantics. As there isn't a `ragel` chat room I can ask for help, as far as I know, I figure that this is the best place to ask.

I've focused on one specific test, `indic_old_spec_003`. The input to the parser looks like this:

```
categories: [C, H, CM, H, X]
```

The parser for the `consonant_syllable` rule looks like this:

```ragel
c = (C | Ra);			# is_consonant
n = ((ZWNJ?.RS)? (N.N?)?);	# is_consonant_modifier
z = ZWJ|ZWNJ;			# is_joiner
reph = (Ra H | Repha);		# possible reph

cn = c.ZWJ?.n?;
symbol = Symbol.N?;
matra_group = z*.(M | SM? MPst).N?.H?;
syllable_tail = (z?.SM.SM?.ZWNJ?)? (A | VD)*;
halant_group = (z?.H.(ZWJ.N?)?);
final_halant_group = halant_group | H.ZWNJ;
medial_group = CM?;
halant_or_matra_group = (final_halant_group | matra_group*);

complex_syllable_tail = (halant_group.cn)* medial_group halant_or_matra_group syllable_tail;
consonant_syllable =	(Repha|CS)? cn complex_syllable_tail;
broken_cluster =	reph? n? complex_syllable_tail;
other =			any;
```

[Source](https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-ot-shaper-indic-machine.rl)

For this rule, the first `(Repha|CS)?` block evaluates to no input, as the first item is `C` which is neither `Repha` nor `CS`. The next item is `cn`, which matches the `C` tag and consumed it. The next tag is `H`, which doesn't match `ZWJ` or `n`, so the `cn` rule completes and we move on to `complex_syllable_tail`.

The `complex_syllable_tail` rule starts with `(halant_group.cn)*`, which would originally match the `H`, but there is no `C` at the end, so this rule evaluates to no input. the next `medial_group` rule is `CM?`. As the current input is `H`, `CM` doesn't match, so this rule evaluates to no input as well. `halant_or_matra_group` goes to `final_halant_group` which goes to `halant_group` which matches `H` and nothing else, consuming it. Finally, `syllable_tail` matches the lack of input at the end. Therefore the range from `0..2` is classified as a constant syllable.

Then, the next item on the chopping block is `CM`. Out of all the rules, this matches the `complex_syllable_tail` part of `broken_cluster`, along with the `H`. Therefore `2..4` is a broken cluster. Finally, the `X` at the end becomes a non-indic character.

However, this fails the test. After wiring some telemetry to the current `rustybuzz` master, I've found that it classifies the range from `0..4` as a consonant syllable and the `4..5` range as non-indic. I'm not sure how it does this; it feels like what would happen is that the `H` is somehow consumed by either the `cn` or the `halant_group.cn` before the `CM` is consumed by the `medial_group`. Unless `ragel`'s semantics are wildly different from what I understand, it is unclear to me how this would happen.

Is there anyone who knows `ragel` well enough to help my understanding of it here?

cc #74, @bluebear94

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ unicode-ccc = "0.1.2"
 unicode-properties = { version = "0.1.0", default-features = false, features = ["general-category"] }
 unicode-script = "0.5.2"
 libm = { version = "0.2.2", optional = true }
+nom = { version = "7.1.3", default-features = false, features = ["alloc"] }
 
 [dependencies.ttf-parser]
 version = "0.19"
--- a/src/complex/indic_machine.rs
+++ b/src/complex/indic_machine.rs
@@ -1,348 +1,251 @@
 use crate::buffer::Buffer;
+use crate::scan::scan;
 
-const MACHINE_TRANS_KEYS: &[u8] = &[
-    8, 8, 4, 8, 5, 7, 5, 8, 4, 8, 6, 6, 16, 16, 4, 8,
-    4, 13, 4, 8, 8, 8, 5, 7, 5, 8, 4, 8, 6, 6, 16, 16,
-    4, 8, 4, 13, 4, 13, 4, 13, 8, 8, 5, 7, 5, 8, 4, 8,
-    6, 6, 16, 16, 4, 8, 4, 8, 4, 13, 8, 8, 5, 7, 5, 8,
-    4, 8, 6, 6, 16, 16, 4, 8, 4, 8, 5, 8, 8, 8, 1, 19,
-    3, 17, 3, 17, 4, 17, 1, 16, 5, 10, 5, 10, 10, 10, 5, 10,
-    1, 16, 1, 16, 1, 16, 3, 10, 4, 10, 5, 10, 4, 10, 5, 10,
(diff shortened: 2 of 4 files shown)


RazrFalcon:
I'm not sure I would call the `nom` variant more maintainable. And performance is a big question as well.

If you have time, I would suggest compiling master `ragel` (which is an `autotools` hell and you need [colm](https://github.com/adrian-thurston/colm) as well) and try its Rust output generator. I've tried it a couple of weeks ago and it was failing with a cryptic error. 
But I think that modifying hb ragel files to output Rust code directly via ragel is far better from maintainability standpoint.

notgull:
Superseded by #77

PR #78, opened 2023-08-25
Add a codegen check to the CI

notgull:
This ensures that all generated code is up to date.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,8 @@ on: [push, pull_request]
 
 env:
   CARGO_TERM_COLOR: always
+  COLM_HASH: 28b6e0a01157049b4cb279b0ef25ea9dcf3b46ed
+  RAGEL_HASH: 65540b65ff09330b0293423e3fecc44e63f30614
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


RazrFalcon:
Sorry, but I don't think this is a good idea. I don't want to have a complicated CI. 
Especially building C from sources. A minor GCC/autotools update and good luck figuring out what is wrong. Nah.

Also, you cannot really forget to update those files, because tests would be failing.

notgull:
Okay, I'll reopen it once there is progress on https://github.com/taiki-e/install-action/issues/199

RazrFalcon:
No, I don't think we need _any_ changes to CI at all. It's fine.

notgull:
Are you sure? I help maintain [`x11rb`](https://github.com/psychon/x11rb), which uses a lot of pre-generated code. Having a check like this is invaluable for preventing the usual mistake of "forgot to run the build scripts again".

RazrFalcon:
>forgot to run the build scripts again

It would not happen, because tests would fail. We have almost 2000 tests. The chance you forget something is tiny.
Again, simplicity is king. The current CI is good enough.

PR #84, opened 2023-10-16
Update bitflags dependency to 2.4.1

bluebear94:
A smaller PR this time. This will require a major version bump because it changes the API of some public types.

### DIFF
--- a/src/ot/map.rs
+++ b/src/ot/map.rs
@@ -135,7 +135,7 @@ impl Map {
 
 bitflags::bitflags! {
     /// Flags used for serialization with a `BufferSerializer`.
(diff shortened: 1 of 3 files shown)


src/ot/map.rs
```diff
@@ -135,7 +135,7 @@ impl Map {
 
 bitflags::bitflags! {
     /// Flags used for serialization with a `BufferSerializer`.
-    #[derive(Default)]
+    #[derive(Default, Clone, Copy, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
notgull, line 138: Do we really need all of these traits?

PR #85, opened 2023-10-16
Allow setting pre- and post-contexts in `UnicodeBuffer`

### LINKED ISSUE #60: Allow specifying `item_offset` and `item_length` as in HarfBuzz
HarfBuzz’s [`hb_buffer_add_codepoints`](https://harfbuzz.github.io/harfbuzz-hb-buffer.html#hb-buffer-add-codepoints) and friends allow specifying an offset and length from the string that was passed in. This allows HarfBuzz to have the full context of the string being shaped in order to perform proper shaping across runs. I’m not sure how feasible it would be for this crate to implement similar functionality, though.

bluebear94:
Resolves #60.

Only dealing with the UTF-8 version for now. I personally need a UTF-16 version as well, but generalizing this for other encodings can come later.

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -1,5 +1,6 @@
 use alloc::{string::String, vec::Vec};
 use core::convert::TryFrom;
+use core::mem;
 
 use ttf_parser::GlyphId;
 
@@ -1298,6 +1299,10 @@ impl Buffer {
         self.idx += count;
     }
 
+    fn clear_context(&mut self, side: usize) {
+        self.context_len[side] = 0;
+    }
+
     pub fn sort(&mut self, start: usize, end: usize, cmp: impl Fn(&GlyphInfo, &GlyphInfo) -> bool) {
         assert!(!self.have_positions);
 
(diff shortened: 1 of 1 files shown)


src/buffer.rs
```diff
@@ -1372,11 +1377,44 @@ impl Buffer {
...
     }
 
-    fn push_str(&mut self, text: &str) {
-        self.ensure(self.len + text.chars().count());
+    fn add_utf8(&mut self, text: &str, item_offset: usize, item_length: usize) {
+        if item_length > (i32::MAX / 8) as usize {
+            return;
+        }
+
+        self.ensure(self.len + item_length * mem::size_of::<u8>() / 4);
```
RazrFalcon, line 1385: Do we really need size_of for `u8`? It's always 1.

src/buffer.rs
```diff
@@ -1580,7 +1618,14 @@ impl UnicodeBuffer {
...
     #[inline]
     pub fn push_str(&mut self, str: &str) {
-        self.0.push_str(str);
+        self.0.add_utf8(str, 0, str.len());
+    }
+
+    /// Pushes a substring of a larger string to a buffer.
+    /// The rest of the string acts as context for shaping.
+    #[inline]
+    pub fn add_utf8(&mut self, str: &str, item_offset: usize, item_length: usize) {
```
RazrFalcon, line 1627: Shouldn't we simply accept 3 `&str`?

src/buffer.rs
bluebear94: Mostly to follow [this part](https://github.com/harfbuzz/harfbuzz/blob/430224b1a06c6eef04de9b6e1ee5aa2cb7a9a403/src/hb-buffer.cc#L1521).

src/buffer.rs
bluebear94: Good point. The signature above follows HarfBuzz’s API, but I think accepting 3 `&str`s is more elegant as a solution. It at least removes the possibility of panicking if `item_offset` and `item_offset + item_length` don’t lie on UTF-8 boundaries.

src/buffer.rs
RazrFalcon: I understand, but we don't have to preserve C language limitations.

src/buffer.rs
bluebear94: Come to think of it, we could alternatively add separate methods for adding pre- and post-context. I’m not sure which design is better.

RazrFalcon: Yes, `set_pre_context` and `set_post_context` is a better choice.

src/buffer.rs
```diff
@@ -1,5 +1,6 @@
 use alloc::{string::String, vec::Vec};
 use core::convert::TryFrom;
+use core::mem;
```
RazrFalcon, line 3: No longer needed.

RazrFalcon:
Thanks!

khaledhosny:
FWIW, HarfBuzz API is designed so that when you are shaping a piece of text, e.g. a paragraph, you can itemize it then shape each item by passing the whole string and the item offset and length without having to split the string.

RazrFalcon:
I understand. Lucky in Rust it's a very simple and safe operation.

PR #88, opened 2023-11-28
Add support for reusable ShapePlans.

vorporeal:
This exposes `ShapePlan` outside of the `rustybuzz` crate, and provides a new `shape_with_plan` function that takes in a `ShapePlan` instead of a list of user features.

Callers are expected to properly cache `ShapePlan`s to avoid incorrect rendering due to mismatches between the `ShapePlan` and the provided buffer.  To help avoid bugs due to mismatches, I added two `debug_assert!()` lines to verify that the plan and buffer use the same direction and script.

Given that `ShapePlan` is now public outside of the crate, I changed a number of `pub` specifiers to `pub(crate)` to maintain the same effective visibility.

### DIFF
--- a/src/plan.rs
+++ b/src/plan.rs
@@ -1,44 +1,50 @@
 use alloc::boxed::Box;
+use alloc::vec::Vec;
 use core::any::Any;
 
 use crate::complex::{complex_categorize, ComplexShaper, DEFAULT_SHAPER, DUMBER_SHAPER};
 use crate::ot::{self, feature, FeatureFlags, TableIndex};
 use crate::{aat, Direction, Face, Feature, Language, Mask, Script, Tag};
 
+/// A reusable plan for shaping a text buffer.
 pub struct ShapePlan {
-    pub direction: Direction,
-    pub script: Option<Script>,
-    pub shaper: &'static ComplexShaper,
-    pub ot_map: ot::Map,
-    pub aat_map: aat::Map,
+    pub(crate) direction: Direction,
+    pub(crate) script: Option<Script>,
+    pub(crate) shaper: &'static ComplexShaper,
+    pub(crate) ot_map: ot::Map,
+    pub(crate) aat_map: aat::Map,
     data: Option<Box<dyn Any>>,
 
-    pub frac_mask: Mask,
-    pub numr_mask: Mask,
-    pub dnom_mask: Mask,
-    pub rtlm_mask: Mask,
-    pub kern_mask: Mask,
-    pub trak_mask: Mask,
+    pub(crate) frac_mask: Mask,
+    pub(crate) numr_mask: Mask,
+    pub(crate) dnom_mask: Mask,
+    pub(crate) rtlm_mask: Mask,
+    pub(crate) kern_mask: Mask,
+    pub(crate) trak_mask: Mask,
 
-    pub requested_kerning: bool,
-    pub requested_tracking: bool,
-    pub has_frac: bool,
-    pub has_vert: bool,
-    pub has_gpos_mark: bool,
-    pub zero_marks: bool,
-    pub fallback_glyph_classes: bool,
-    pub fallback_mark_positioning: bool,
-    pub adjust_mark_positioning_when_zeroing: bool,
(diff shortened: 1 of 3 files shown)


src/shape.rs
```diff
@@ -7,32 +7,57 @@ use crate::buffer::{
...
-            face,
-            buffer.direction,
-            buffer.script,
-            buffer.language.as_ref(),
-            features,
-        );
+    debug_assert_eq!(buffer.direction, plan.direction);
+    debug_assert_eq!(
+        buffer.script.unwrap_or(script::UNKNOWN),
+        plan.script.unwrap_or(script::UNKNOWN)
```
vorporeal, line 51: The need for `.unwrap_or(script::UNKNOWN)` here is due to the fact that the `UnicodeBuffer::script()` function returns a `Script` (instead of an `Option<Script>`), and performs the same unwrapping logic.

The shape plan caching I implemented in `cosmic-text` to test this invoked `UnicodeBuffer::script()` to get the script value to pass to the `ShapePlan` constructor, so the assertion would fail if `buffer.0.script` is `None`, because the `ShapePlan` would have been constructed with `Some(script::UNKNOWN)`.

src/shape.rs
```diff
@@ -7,32 +7,57 @@ use crate::buffer::{
...
+///
+/// It is up to the caller to ensure that the shape plan matches the properties of the provided
+/// buffer, otherwise the shaping result will likely be incorrect.
+///
+/// # Panics
+///
+/// Will panic when debugging assertions are enabled if the buffer and plan have mismatched
+/// properties.
+pub fn shape_with_plan(face: &Face, plan: &ShapePlan, mut buffer: UnicodeBuffer) -> GlyphBuffer {
+    buffer.guess_segment_properties();
```
RazrFalcon, line 43: Why twice?

vorporeal: Ah sorry was testing out an alternative approach to avoid the `.unwrap_or(script::UNKNOWN)` awkwardness below and apparently forgot to remove this line.  Removed.

src/plan.rs
```diff
@@ -67,6 +73,7 @@ pub struct ShapePlanner<'a> {
     pub script_zero_marks: bool,
     pub script_fallback_mark_positioning: bool,
     pub shaper: &'static ComplexShaper,
+    pub user_features: Vec<Feature>,
```
RazrFalcon, line 76: Why do we have to allocate it again?

src/shape.rs
```diff
@@ -7,32 +7,55 @@ use crate::buffer::{
...
+use crate::{
+    aat, fallback, normalize, ot, script, Direction, Face, Feature, GlyphBuffer, UnicodeBuffer,
+};
 
 /// Shapes the buffer content using provided font and features.
 ///
 /// Consumes the buffer. You can then run `GlyphBuffer::clear` to get the `UnicodeBuffer` back
 /// without allocating a new one.
-pub fn shape(face: &Face, features: &[Feature], buffer: UnicodeBuffer) -> GlyphBuffer {
+pub fn shape(face: &Face, features: &[Feature], mut buffer: UnicodeBuffer) -> GlyphBuffer {
```
RazrFalcon, line 18: I would add something along the lines to the doc here:

```
If you plan to shape multiple strings using the same [`Face`] prefer [`shape_with_plan`].
This is because [`ShapePlan`] initialization is pretty slow and should preferably be called once for each [`Face`].
```

src/plan.rs
RazrFalcon: But `compile` is a private method. It cannot be misused.

As for allocation, yes, my bad. There is none. But the code is still confusing. It confused me. I would prefer to avoid storing `user_features` in the `ShapePlanner`.

vorporeal: Sure, done.  I'm used to working on codebases with many more contributors (and PR approvers), where there's more value in protecting even private functions from misuse within the codebase. :)

src/shape.rs
vorporeal: Good call; done.

RazrFalcon:
Thanks! I will publish a new version soon.

PR #90, opened 2023-12-06
Make shape plan `Send` and `Sync`

laurmaedje:
With this trivial change a single shape plan can be used from multiple threads.


laurmaedje:
Thanks for the quick merges!

YevheniiReizner:
Are you fine using git version or do you need a new release?

RazrFalcon:
Are you fine using git version or do you need a new release?

laurmaedje:
A new release would be great, but I'm not in a hurry, so take your time.

PR #95, opened 2024-02-09
3.3.2

LaurenzV:
| Status | Commit message | HB | Comments |
| ------ | -------------------------------------------------------------------------------------- | ------------------------------------------------------------- | --------------- |
| ⚪️ | [ci] Try to fix the randomly failing valgrind job | [Link](https://github.com/harfbuzz/harfbuzz/commit/52a33a15a) | |
| ⚪️ | [ci] Try harder to fix this randomly failing job | [Link](https://github.com/harfbuzz/harfbuzz/commit/449c4296a) | |
| ⚪️ | [repacker] don't infinite loop if visited or roots is in error. | [Link](https://github.com/harfbuzz/harfbuzz/commit/c4573c2ec) | |
| ⚪️ | [subset] Fix for fuzzer timeout. | [Link](https://github.com/harfbuzz/harfbuzz/commit/067f90a82) | |
| ⚪️ | [font] Add public API for slant | [Link](https://github.com/harfbuzz/harfbuzz/commit/810f5d71c) | |
| ⚪️ | [GPOS] Apply font synthetic slant | [Link](https://github.com/harfbuzz/harfbuzz/commit/ff697debd) | |
| ⚪️ | [style] Adjust font slant angle for synthetic slant value | [Link](https://github.com/harfbuzz/harfbuzz/commit/602bd3968) | |
| 🟢 | Use invisible-glyph for spaces if font has no ASCII space | [Link](https://github.com/harfbuzz/harfbuzz/commit/06ee40213) | |
| ⚪️ | [hb-ot-var] Actually set in/out argument | [Link](https://github.com/harfbuzz/harfbuzz/commit/1b51be57d) | |
| ⚪️ | [hb-ot-var] Specify normalized 2.14 docs | [Link](https://github.com/harfbuzz/harfbuzz/commit/6da4b80e5) | |
| ⚪️ | [face] Clarify face_index handling | [Link](https://github.com/harfbuzz/harfbuzz/commit/da7dba000) | |
| ⚪️ | [font] Load named-instance if face index top bits are set | [Link](https://github.com/harfbuzz/harfbuzz/commit/2d42fc9fb) | |
| ⚪️ | [font] Mark hb_font_get_var_coords_design() non-experimental | [Link](https://github.com/harfbuzz/harfbuzz/commit/5b87c30d6) | |
| ⚪️ | [docs] Update | [Link](https://github.com/harfbuzz/harfbuzz/commit/6356da88e) | |
| ⚪️ | Fix test | [Link](https://github.com/harfbuzz/harfbuzz/commit/0df83f52a) | |
| ⚪️ | Typo | [Link](https://github.com/harfbuzz/harfbuzz/commit/a3265e1e0) | |
| ⚪️ | [metrics] Scale up horizontal caret rise/run | [Link](https://github.com/harfbuzz/harfbuzz/commit/fc329f665) | |
| ⚪️ | [metrics] Implement synthetic slant for caret slope | [Link](https://github.com/harfbuzz/harfbuzz/commit/fa008a021) | |
| ⚪️ | [metrics] Harden math | [Link](https://github.com/harfbuzz/harfbuzz/commit/45c2b6e38) | |
| ⚪️ | [metrics] Only scale caret rise/run if font is slanted | [Link](https://github.com/harfbuzz/harfbuzz/commit/f1ba1f940) | |
| ⚪️ | [metrics] Simplify | [Link](https://github.com/harfbuzz/harfbuzz/commit/f7d49d9c2) | |
| ⚪️ | [metrics] Ouch. Fix slant code | [Link](https://github.com/harfbuzz/harfbuzz/commit/4ab94443b) | |
| ⚪️ | [docs] Add hb_font_[gs]et_synthetic_slant() | [Link](https://github.com/harfbuzz/harfbuzz/commit/a7751277e) | |
| ⚪️ | [metrics] Fix slant calc | [Link](https://github.com/harfbuzz/harfbuzz/commit/3fc2e654c) | |
| ⚪️ | Merge pull request #3338 from harfbuzz/slant | [Link](https://github.com/harfbuzz/harfbuzz/commit/165a6073f) | |
| 🟢 | Add hb_segment_properties_overlay() | [Link](https://github.com/harfbuzz/harfbuzz/commit/6a6699284) | |
| 🟢 | [buffer] Overlay segment-properties in hb_buffer_append() | [Link](https://github.com/harfbuzz/harfbuzz/commit/36e9f467a) | |
| ⚪️ | [util] Move copy_buffer_properties() out of loop | [Link](https://github.com/harfbuzz/harfbuzz/commit/101d07efd) | |
| ⚪️ | [util] Simplify copy_buffer_properties() | [Link](https://github.com/harfbuzz/harfbuzz/commit/34f5b855a) | |
| ⚪️ | [util] Copy unicode_funcs in copy_buffer_properties() | [Link](https://github.com/harfbuzz/harfbuzz/commit/d0c3515ce) | |
| 🟢 | [buffer] Clean up internal state bookkeeping | [Link](https://github.com/harfbuzz/harfbuzz/commit/f643b81ff) | |
| ⚪️ | [buffer] Add hb_buffer_create_similar() | [Link](https://github.com/harfbuzz/harfbuzz/commit/52f5711ed) | |
| 🟢 | [buffer] Group shape-related members together | [Link](https://github.com/harfbuzz/harfbuzz/commit/43be5ba44) | |
| 🟢 | [buffer] Add enter()/leave() pair around shape() | [Link](https://github.com/harfbuzz/harfbuzz/commit/d1e7df5c3) | |
| ⚪️ | Merge pull request #3353 from harfbuzz/buffer-create-similar | [Link](https://github.com/harfbuzz/harfbuzz/commit/c5e5d5e0b) | |
| 🟢 | [buffer] Rename swap_buffers() to sync() | [Link](https://github.com/harfbuzz/harfbuzz/commit/bea5369c6) | |
| ⚪️ | [failing-alloc] Make it compile as C++ as well | [Link](https://github.com/harfbuzz/harfbuzz/commit/0a5c1c970) | |
| ⚪️ | docs: Add some details about coordinates | [Link](https://github.com/harfbuzz/harfbuzz/commit/d70825a6d) | |
| ⚪️ | Fix unintentional locale dependency (#3358) | [Link](https://github.com/harfbuzz/harfbuzz/commit/a45a63053) | |
| ⚪️ | [map] Actually use k/v invalid types for declaration! | [Link](https://github.com/harfbuzz/harfbuzz/commit/6f559346a) | |
| ⚪️ | Revert "[map] Actually use k/v invalid types for declaration!" | [Link](https://github.com/harfbuzz/harfbuzz/commit/ce7f19a2c) | |
| ⚪️ | Use freetype from CMake target when present (#3361) | [Link](https://github.com/harfbuzz/harfbuzz/commit/b8c2c1ab3) | |
| ⚪️ | [map] Map == / != use correct types | [Link](https://github.com/harfbuzz/harfbuzz/commit/74811c501) | |
| ⚪️ | [algs] Call std::hash from hb_hash() | [Link](https://github.com/harfbuzz/harfbuzz/commit/53847dba9) | |
| ⚪️ | Clean up HB_NO_SETLOCALE | [Link](https://github.com/harfbuzz/harfbuzz/commit/13643934a) | |
| ⚪️ | color: Document empty returns (#3362) | [Link](https://github.com/harfbuzz/harfbuzz/commit/befe3b5d5) | |
| ⚪️ | docs: Clarify variation apis (#3363) | [Link](https://github.com/harfbuzz/harfbuzz/commit/72f0b24ee) | |
| ⚪️ | Fix compiler warning | [Link](https://github.com/harfbuzz/harfbuzz/commit/0a42f46ce) | |
| ⚪️ | More macro cleanup | [Link](https://github.com/harfbuzz/harfbuzz/commit/0b2c514fb) | |
| ⚪️ | Preserve glyph props | [Link](https://github.com/harfbuzz/harfbuzz/commit/96c969c82) | |
| ⚪️ | Rename variable | [Link](https://github.com/harfbuzz/harfbuzz/commit/9408e0d6d) | |
| ⚪️ | Rename method | [Link](https://github.com/harfbuzz/harfbuzz/commit/fa12f1ae2) | |
| ⚪️ | [doc] Fix hb_font_set_synthetic_slant param name | [Link](https://github.com/harfbuzz/harfbuzz/commit/0acf466c4) | |
| ⚪️ | [git.mk] Update | [Link](https://github.com/harfbuzz/harfbuzz/commit/16cfe34ae) | |
| ⚪️ | docs: Fix a typo | [Link](https://github.com/harfbuzz/harfbuzz/commit/bb330c6aa) | |
| ⚪️ | docs: Add some details | [Link](https://github.com/harfbuzz/harfbuzz/commit/4ced2e8a8) | |
| ⚪️ | [subset] fix fuzzer timeout if visisted_paint goes into error. | [Link](https://github.com/harfbuzz/harfbuzz/commit/87496bf63) | |
| ⚪️ | [algs] Fix hash chaining to std::hash() | [Link](https://github.com/harfbuzz/harfbuzz/commit/cfe2b6611) | |
| ⚪️ | [map] Massage some more | [Link](https://github.com/harfbuzz/harfbuzz/commit/ca4f56bdc) | |
| ⚪️ | Fix the docs build | [Link](https://github.com/harfbuzz/harfbuzz/commit/b15e497d4) | |
| ⚪️ | [map] Construct objects | [Link](https://github.com/harfbuzz/harfbuzz/commit/114046dc2) | |
| ⚪️ | [map] Remove constexpr invalid items | [Link](https://github.com/harfbuzz/harfbuzz/commit/5a91db113) | |
| ⚪️ | [algs] Add hb_coerce() | [Link](https://github.com/harfbuzz/harfbuzz/commit/1e27f95dc) | |
| ⚪️ | [check-static-inits] Only check library object files | [Link](https://github.com/harfbuzz/harfbuzz/commit/98b26eedf) | |
| ⚪️ | [map] Allow invalid items to be pointer to static object | [Link](https://github.com/harfbuzz/harfbuzz/commit/985b63b3e) | |
| ⚪️ | [test-map] Add disabled tests with std::string | [Link](https://github.com/harfbuzz/harfbuzz/commit/bc0a5fdf4) | |
| ⚪️ | Convert fallback kwargs to [provide] entries. | [Link](https://github.com/harfbuzz/harfbuzz/commit/726b1a8b2) | |
| ⚪️ | [map] Destruct objects | [Link](https://github.com/harfbuzz/harfbuzz/commit/956e0a4d1) | |
| ⚪️ | [meta] Replace hb_is_pointer with std::is_pointer | [Link](https://github.com/harfbuzz/harfbuzz/commit/47afb3ef0) | |
| ⚪️ | [meta] Use std::is_reference instead of hb_is_reference | [Link](https://github.com/harfbuzz/harfbuzz/commit/909dde9df) | |
| ⚪️ | [meta] Use std::is_const instead of hb_is_const | [Link](https://github.com/harfbuzz/harfbuzz/commit/3b2e60423) | |
| ⚪️ | [meta] Remove hb_add_const | [Link](https://github.com/harfbuzz/harfbuzz/commit/b7b0a15f7) | |
| ⚪️ | [meta] Use std::addressof() instead of hb_addressof() | [Link](https://github.com/harfbuzz/harfbuzz/commit/8a69e0063) | |
| ⚪️ | [map] Fix bad memory access if hb_map.fini() was called twice. | [Link](https://github.com/harfbuzz/harfbuzz/commit/ff4e8c7ea) | |
| ⚪️ | [map] Correct previous commit | [Link](https://github.com/harfbuzz/harfbuzz/commit/e30511a8d) | |
| ⚪️ | [meta] Remove unused hb_ref() | [Link](https://github.com/harfbuzz/harfbuzz/commit/71e6f742a) | |
| ⚪️ | [meta] Include <memory> | [Link](https://github.com/harfbuzz/harfbuzz/commit/1c5010660) | |
| ⚪️ | [vector] Move semantics when resizing | [Link](https://github.com/harfbuzz/harfbuzz/commit/63affc4ea) | |
| ⚪️ | Further adjust setlocale | [Link](https://github.com/harfbuzz/harfbuzz/commit/b97e4f708) | |
| ⚪️ | Fix previous commit | [Link](https://github.com/harfbuzz/harfbuzz/commit/589bea15b) | |
| 🟢 | Fix various typos | [Link](https://github.com/harfbuzz/harfbuzz/commit/e2e305066) | |
| ⚪️ | [algs] Add default-construtor to hb_pair_t | [Link](https://github.com/harfbuzz/harfbuzz/commit/c0f57ccc8) | |
| ⚪️ | Try fix Mac build | [Link](https://github.com/harfbuzz/harfbuzz/commit/1173ed1f7) | |
| ⚪️ | Better try at previous commit | [Link](https://github.com/harfbuzz/harfbuzz/commit/19793f333) | |
| ⚪️ | Make hb_coerce static inline | [Link](https://github.com/harfbuzz/harfbuzz/commit/92f179075) | |
| ⚪️ | [vector] Add XXX markers for remaining places that need work | [Link](https://github.com/harfbuzz/harfbuzz/commit/65a22836f) | |
| ⚪️ | [vector] Start adding destruction | [Link](https://github.com/harfbuzz/harfbuzz/commit/813eaba6e) | |
| ⚪️ | [vector] Construct items when enlarging | [Link](https://github.com/harfbuzz/harfbuzz/commit/7171917b5) | |
| ⚪️ | [vector] Destruct in pop() | [Link](https://github.com/harfbuzz/harfbuzz/commit/5946e945d) | |
| ⚪️ | [vector] Move semantics in vector remove() | [Link](https://github.com/harfbuzz/harfbuzz/commit/c58bfa35f) | |
| ⚪️ | [vector] Adjust construction criteria | [Link](https://github.com/harfbuzz/harfbuzz/commit/1198fb246) | |
| ⚪️ | [repacker] Replace fini_deep() with fini() | [Link](https://github.com/harfbuzz/harfbuzz/commit/15cceff3d) | |
| ⚪️ | Merge pull request #3376 from harfbuzz/auto-vector | [Link](https://github.com/harfbuzz/harfbuzz/commit/0b54f9241) | |
| ⚪️ | [cff] Remove unneeded init/fini | [Link](https://github.com/harfbuzz/harfbuzz/commit/c3dd21f11) | |
| ⚪️ | [bimap] Remove init/fini | [Link](https://github.com/harfbuzz/harfbuzz/commit/de2cef42b) | |
| ⚪️ | [str_buff_vec_t] Remove unused fini method | [Link](https://github.com/harfbuzz/harfbuzz/commit/035f03346) | |
| ⚪️ | [subset-cff2] Drop an constructor/destructor pair | [Link](https://github.com/harfbuzz/harfbuzz/commit/a1e9a276e) | |
| ⚪️ | [subset-cff1] Remove a constructor/destructor pair | [Link](https://github.com/harfbuzz/harfbuzz/commit/07474aa8c) | |
| ⚪️ | [subset-cff] Remove another set of fini_deep | [Link](https://github.com/harfbuzz/harfbuzz/commit/a48aa9850) | |
| ⚪️ | [subset-cff] Convert subr_closures_t to constructor/destructor instead of init/fini. | [Link](https://github.com/harfbuzz/harfbuzz/commit/f1dc8b087) | |
| ⚪️ | [vector] Add TODO Emplace? | [Link](https://github.com/harfbuzz/harfbuzz/commit/213e3f09b) | |
| ⚪️ | [cff] Remove init/fini from blend_arg_t | [Link](https://github.com/harfbuzz/harfbuzz/commit/29f60d1ea) | |
| ⚪️ | [cff] Remove some more fini_deep() | [Link](https://github.com/harfbuzz/harfbuzz/commit/4eb6e6ea1) | |
| ⚪️ | [meson] add icu DEFS required for compilation | [Link](https://github.com/harfbuzz/harfbuzz/commit/daab4bf63) | |
| ⚪️ | [vector] Remove .fini_deep() | [Link](https://github.com/harfbuzz/harfbuzz/commit/6ef83c2c3) | |
| ⚪️ | [cff] Remove init/fini from number_t | [Link](https://github.com/harfbuzz/harfbuzz/commit/8d24ac266) | |
| ⚪️ | [cff] Remove op_str_t nop init/fini | [Link](https://github.com/harfbuzz/harfbuzz/commit/fe783ac09) | |
| ⚪️ | Merge pull request #3381 from harfbuzz/clean-vector-use | [Link](https://github.com/harfbuzz/harfbuzz/commit/10a20be39) | |
| ⚪️ | [subset] Fix bound check when setting overlap bit. | [Link](https://github.com/harfbuzz/harfbuzz/commit/94ba4d5a4) | |
| 🟢 | Add test for #2516 | [Link](https://github.com/harfbuzz/harfbuzz/commit/fc5354c5e) | |
| 🟢 | Test for #2140 | [Link](https://github.com/harfbuzz/harfbuzz/commit/282642311) | |
| ⚪️ | meson: Enable big objects support when building for windows | [Link](https://github.com/harfbuzz/harfbuzz/commit/17cd4457b) | |
| ⚪️ | Merge pull request #3365 from harfbuzz/gdef-fix | [Link](https://github.com/harfbuzz/harfbuzz/commit/c8195dbc0) | |
| 🟢 | [gpos] Fix unsafe-to-break of mark-attachment | [Link](https://github.com/harfbuzz/harfbuzz/commit/7e160dc34) | |
| ⚪️ | [ms-feature-ranges] Use preferred vector search API | [Link](https://github.com/harfbuzz/harfbuzz/commit/849e40b4d) | |
| ⚪️ | [vector] Remove old find() method | [Link](https://github.com/harfbuzz/harfbuzz/commit/cd9799784) | |
| ⚪️ | [vector] Add sorted template argument | [Link](https://github.com/harfbuzz/harfbuzz/commit/294d50ff3) | |
| ⚪️ | [vector] Merge sorted-vector into vector | [Link](https://github.com/harfbuzz/harfbuzz/commit/1,06E+81) | |
| ⚪️ | [coretext] Fix lsearch | [Link](https://github.com/harfbuzz/harfbuzz/commit/376738d41) | |
| ⚪️ | Merge pull request #3386 from harfbuzz/unify-sorted-vector | [Link](https://github.com/harfbuzz/harfbuzz/commit/cfa00238e) | |
| ⚪️ | [ms-feature-ranges] Inline code in header file | [Link](https://github.com/harfbuzz/harfbuzz/commit/ac109c935) | |
| ⚪️ | [ms-feature-ranges] Pass reference to cmp function | [Link](https://github.com/harfbuzz/harfbuzz/commit/2d12fc90a) | |
| ⚪️ | [test] Add --single-par to more places in hb-aots-tester [ci skip] | [Link](https://github.com/harfbuzz/harfbuzz/commit/1522015bb) | |
| ⚪️ | [machinery] Make accelerator lazy-loader call Xinit/Xfini | [Link](https://github.com/harfbuzz/harfbuzz/commit/e062376ef) | |
| ⚪️ | [repacker] Fix missing initilization of obj in vertex_t. | [Link](https://github.com/harfbuzz/harfbuzz/commit/112cb9fe6) | |
| ⚪️ | [machinery] Move accelerators to constructor/destructor | [Link](https://github.com/harfbuzz/harfbuzz/commit/ac1bb3e39) | |
| ⚪️ | Merge pull request #3392 from harfbuzz/auto-accelerators | [Link](https://github.com/harfbuzz/harfbuzz/commit/0ae665939) | |
| ⚪️ | Clean accelerators a bit more | [Link](https://github.com/harfbuzz/harfbuzz/commit/ae9fc7f53) | |
| 🟢 | [PairPos] Split GPOS kerning to both sides (#3235) | [Link](https://github.com/harfbuzz/harfbuzz/commit/84aa1a836) | |
| ⚪️ | Fix failing Mac test for previous commit | [Link](https://github.com/harfbuzz/harfbuzz/commit/070e30e6d) | |
| ⚪️ | One more fix | [Link](https://github.com/harfbuzz/harfbuzz/commit/cab943a01) | |
| ⚪️ | [post] Initialize variables | [Link](https://github.com/harfbuzz/harfbuzz/commit/773cf4fdd) | |
| ⚪️ | [gpos] Fix conditional | [Link](https://github.com/harfbuzz/harfbuzz/commit/cce37a6f1) | |
| ⚪️ | [cff] Initialize accelerator members | [Link](https://github.com/harfbuzz/harfbuzz/commit/a8b7f1880) | |
| 🟢 | [buffer] Add HB_GLYPH_FLAG_UNSAFE_TO_CONCAT | [Link](https://github.com/harfbuzz/harfbuzz/commit/3122c2cdc) | |
| ⚪️ | Implement hb-shape --verify unsafe-to-concat flag | [Link](https://github.com/harfbuzz/harfbuzz/commit/36b156171) | |
| 🟢 | [unsafe-to-concat] Add annotations to GPOS and kern | [Link](https://github.com/harfbuzz/harfbuzz/commit/e1cbd4539) | |
| 🟢 | [unsafe-to-concat] Add to GPOS kerning | [Link](https://github.com/harfbuzz/harfbuzz/commit/596bc7e93) | |
| 🟢 | [gsubgpos] Combine input/backtrack/lookahead unsafe-to-concat | [Link](https://github.com/harfbuzz/harfbuzz/commit/78481b32c) | |
| 🟢 | [buffer] Rename \_unsafe_to_break_set_mask to \_infos_set_glyph_flags | [Link](https://github.com/harfbuzz/harfbuzz/commit/56d081955) | |
| 🟢 | [buffer] Add default cluster value in find_min_cluster | [Link](https://github.com/harfbuzz/harfbuzz/commit/f91ce56e0) | |
| 🟢 | [buffer] Consolidate glyph-flags implementation | [Link](https://github.com/harfbuzz/harfbuzz/commit/d98a0fc88) | |
| 🟢 | [gsubgpos] Adjust chaining unsafe-to-concat application | [Link](https://github.com/harfbuzz/harfbuzz/commit/60006d368) | |
| 🟢 | [unsafe-to-concat] Mark entire buffer unsafe-to-concat if kerx format2 | [Link](https://github.com/harfbuzz/harfbuzz/commit/c0058892b) | |
| 🟢 | [unsafe-to-concat] Fix PairPos2 logic | [Link](https://github.com/harfbuzz/harfbuzz/commit/48c5f2619) | |
| 🟢 | [unsafe-to-concat] Adjust Arabic joining logic | [Link](https://github.com/harfbuzz/harfbuzz/commit/235c3a129) | |
| 🟢 | [unsafe-to-concat] Further adjust Arabic joining logic at boundary | [Link](https://github.com/harfbuzz/harfbuzz/commit/4f04baef1) | |
| 🟢 | [unsafe-to-concat] Adjust "interior"ness of "from_out_buffer" | [Link](https://github.com/harfbuzz/harfbuzz/commit/ea1b32c8c) | |
| 🟢 | [unsafe-to-concat] Adjust CursivePos | [Link](https://github.com/harfbuzz/harfbuzz/commit/11bdd7a02) | |
| 🟢 | [unsafe-to-concat] Adjust Arabic-joining start boundary condition more | [Link](https://github.com/harfbuzz/harfbuzz/commit/909e34f68) | |
| 🟢 | [unsafe-to-concat] Adjust GPOS lookbacks | [Link](https://github.com/harfbuzz/harfbuzz/commit/95d052a79) | |
| 🟢 | [unsafe-to-concat] Adjust MarkBasePos | [Link](https://github.com/harfbuzz/harfbuzz/commit/b443898cc) | |
| 🟢 | [unsafe-to-concat] Mark LigatureSubst | [Link](https://github.com/harfbuzz/harfbuzz/commit/a57599205) | |
| 🟢 | [unsafe-to-concat] More annotations for MarkLigaturePos | [Link](https://github.com/harfbuzz/harfbuzz/commit/6a7d6d4b6) | |
| 🟢 | [unsafe-to-concat] More annotations for MarkMarkPos | [Link](https://github.com/harfbuzz/harfbuzz/commit/8663eda4f) | |
| 🟢 | [unsafe-to-concat] Adjust end conditions | [Link](https://github.com/harfbuzz/harfbuzz/commit/14d43d123) | |
| 🟢 | [unsafe-to-concat] Mark as unsafe in kern machine | [Link](https://github.com/harfbuzz/harfbuzz/commit/bcdfedbc0) | |
| ⚪️ | [fallback-shape] Add buffer trace log | [Link](https://github.com/harfbuzz/harfbuzz/commit/374a6f186) | |
| 🟢 | Cosmetic | [Link](https://github.com/harfbuzz/harfbuzz/commit/6e345f709) | |
| ⚪️ | [unsafe-to-concat] Mark in all other shapers | [Link](https://github.com/harfbuzz/harfbuzz/commit/20031ddbb) | |
| ⚪️ | [buffer] Oops | [Link](https://github.com/harfbuzz/harfbuzz/commit/332460649) | |
| ⚪️ | [test] Add test-serialize | [Link](https://github.com/harfbuzz/harfbuzz/commit/76c6f4164) | |
| ⚪️ | [test-serialize] Assert len | [Link](https://github.com/harfbuzz/harfbuzz/commit/6c09a8b5b) | |
| ⚪️ | [test] Remove HB_UNUSED | [Link](https://github.com/harfbuzz/harfbuzz/commit/bad5c9456) | |
| 🟢 | [unsafe-to-concat] Clarify documentation as per feedback | [Link](https://github.com/harfbuzz/harfbuzz/commit/540af548d) | |
| ⚪️ | Merge pull request #3297 from harfbuzz/unsafe-to-concat | [Link](https://github.com/harfbuzz/harfbuzz/commit/e9cc5f6cd) | |
| ⚪️ | [doc] Fix generation of hb_glyph_flags_t docs | [Link](https://github.com/harfbuzz/harfbuzz/commit/fa29006bf) | |
| ⚪️ | Avoid redefinition of HB_NO_SETLOCALE in certain configs | [Link](https://github.com/harfbuzz/harfbuzz/commit/91b779e23) | |
| ⚪️ | [font] Fix build with no-var configs | [Link](https://github.com/harfbuzz/harfbuzz/commit/24650624c) | |
| ⚪️ | [buffer] Make hb_buffer_append() take a const argument | [Link](https://github.com/harfbuzz/harfbuzz/commit/7c704d898) | |
| ⚪️ | [buffer] Add HB_BUFFER_FLAG_VERIFY | [Link](https://github.com/harfbuzz/harfbuzz/commit/61823838f) | |
| ⚪️ | [config] Enable HB_NO_BUFFER_VERIFY in HB_LEAN | [Link](https://github.com/harfbuzz/harfbuzz/commit/5b1d813b6) | |
| ⚪️ | [fuzz] Verify shape results | [Link](https://github.com/harfbuzz/harfbuzz/commit/6596e42d1) | |
| ⚪️ | [fuzz] Disable verification for now. | [Link](https://github.com/harfbuzz/harfbuzz/commit/61856359c) | |
| ⚪️ | [buffer] Document HB_BUFFER_FLAG_VERIFY | [Link](https://github.com/harfbuzz/harfbuzz/commit/476a6377a) | |
| ⚪️ | [subset] convert active_glyphs_stack to be a vector of hb_set_t instead of hb_set_t\*. | [Link](https://github.com/harfbuzz/harfbuzz/commit/f3c1f4f0d) | |
| ⚪️ | [subset] Fix for issue #3397. | [Link](https://github.com/harfbuzz/harfbuzz/commit/bc899650c) | |
| ⚪️ | Add the language system tag INUK | [Link](https://github.com/harfbuzz/harfbuzz/commit/5a6545940) | already existed |
| 🟢 | Infer tag mappings for unregistered macrolanguages | [Link](https://github.com/harfbuzz/harfbuzz/commit/0e31595e0) | |
| ⚪️ | Replace “[family]” with “[collection]” | [Link](https://github.com/harfbuzz/harfbuzz/commit/0b1bf89cc) | |
| ⚪️ | [buffer] Whitespace | [Link](https://github.com/harfbuzz/harfbuzz/commit/3972e0a8f) | |
| ⚪️ | [util] Change "All shapers failed." message to "Shaping failed." | [Link](https://github.com/harfbuzz/harfbuzz/commit/d35f38012) | |
| ⚪️ | [verify] Show buffer input text when verification fails | [Link](https://github.com/harfbuzz/harfbuzz/commit/e986c1207) | |
| ⚪️ | [fallback-kern] Move buffer message to correct position | [Link](https://github.com/harfbuzz/harfbuzz/commit/b47b3b997) | |
| 🟢 | Don't always inherit from macrolanguages | [Link](https://github.com/harfbuzz/harfbuzz/commit/a184c5f85) | |
| 🟢 | Let BCP 47 tag "mo" fall back to OT tag 'ROM ' | [Link](https://github.com/harfbuzz/harfbuzz/commit/ae9afd977) | |
| ⚪️ | Merge pull request #3402 from harfbuzz/language-tags | [Link](https://github.com/harfbuzz/harfbuzz/commit/3e8742e37) | |
| ⚪️ | 3.3.0 | [Link](https://github.com/harfbuzz/harfbuzz/commit/98da33063) | |
| ⚪️ | [subset] Don't hold references to members of the active_glyph_stack. | [Link](https://github.com/harfbuzz/harfbuzz/commit/4e2f409bc) | |
| ⚪️ | 3.3.1 | [Link](https://github.com/harfbuzz/harfbuzz/commit/45df25953) | |
| ⚪️ | [serialize] document how the serializer works. | [Link](https://github.com/harfbuzz/harfbuzz/commit/0fb8e2240) | |
| ⚪️ | [glyf] Don't store face in accelerator | [Link](https://github.com/harfbuzz/harfbuzz/commit/e59da2871) | |
| 🟢 | [buffer] Comment | [Link](https://github.com/harfbuzz/harfbuzz/commit/8670ffbf4) | |
| ⚪️ | [atexit] Allow hb_atexit redefinition | [Link](https://github.com/harfbuzz/harfbuzz/commit/917795372) | |
| 🟢 | [GPOS] Disable split-kerning | [Link](https://github.com/harfbuzz/harfbuzz/commit/88798ee8b) | |
| ⚪️ | Revert "One more fix" | [Link](https://github.com/harfbuzz/harfbuzz/commit/44da4e2a2) | |
| ⚪️ | Revert "Fix failing Mac test for previous commit" | [Link](https://github.com/harfbuzz/harfbuzz/commit/7657bdd91) | |
| ⚪️ | 3.3.2 | [Link](https://github.com/harfbuzz/harfbuzz/commit/ac46c3248) |


LaurenzV:
Currently stuck at https://github.com/harfbuzz/harfbuzz/commit/84aa1a836. It seems to me like the code we currently have is somewhat different and I also don't understand what's going on there. I think I will have to start familiarize myself with how GPOS works instead of trying to brainlessly copy it, so it might be a while before I make progress on that.

LaurenzV:
Or well, I guess I do understand what's happening more or less, but some of the parts I'm unsure what their equivalent in Rust is. For example the `get_len` method of `ValueFormat`.

RazrFalcon:
We're talking about `src/ot/position.rs PairAdjustment::apply`, right?

LaurenzV:
Yes, I think that's the corresponding code.

RazrFalcon:
If you look at git blame, I'm not the one who wrote that code 😄 
So I guess we have to figure out it together.

RazrFalcon:
As for `ValueFormat::get_len`, this is `ValueFormatFlags::size` in Rust.

RazrFalcon:
But it's private API...

LaurenzV:
Ah! I didn't check `ttf-parser`...

LaurenzV:
Okay, that helps already though, thanks. I'll still do some more background reading though and then give it another shot.

RazrFalcon:
Basically, in harfbuzz we have a mix of shaping and font parsing at the same time here. But in rustybuzz it's all hidden it ttf-parser. And it works quite well. 

But now we have to figure out if this change could be done on the rustybuzz part of should implemented in ttf-parser first. I will take a look into it more to say for sure. 

Don't worry, this part is known to be very complicated. And there is not much we can do about it. It's just inherent complexity of TrueType.

RazrFalcon:
PS: the weird `// Note the intentional use of "|" instead of short-circuit "||".` comment is outdated. We can use `||` for flags here.

LaurenzV:
> PS: the weird // Note the intentional use of "|" instead of short-circuit "||". comment is outdated. We can use || for flags here.

Yes, I already fixed it locally. I had forgotten to remove it in the original commit.

RazrFalcon:
As for:

```c
if (valueFormat1 & !mask)
	  goto bail;
```

This should be:

```rust
if records.0.x_advance == 0 && records.0.x_advance_device.is_none() {
    // goto bail;
}
```

Typical C-style bitflags magic...

And similar logic for Y-axis.

RazrFalcon:
And the other stuff seems to be pretty straight-forward.

`valueFormat1.apply_value()` is `records.0.apply()`

RazrFalcon:
What harfbuzz doing here is parsing [ValueRecord](https://learn.microsoft.com/en-us/typography/opentype/spec/gpos#value-record), but ttf-parser already parsed it. This structure is a bit insane. It's a dynamically sized collection of data of various types. Pretty absurd even by TrueType standards.

LaurenzV:
Alright, took me many hours, but I think I got it now. Not very pretty, but I don't see any better way to translate the goto statements and struct offset magic. Will push it (and prettify it) once I'm done with the remaining commits.

RazrFalcon:
If it takes too much time and effort - just ask and I will try to do it myself.  I don't think it would be much faster, most of the code I wrote was 3 years ago, but I don't want you to burn out immediately.

LaurenzV:
https://github.com/harfbuzz/harfbuzz/commit/88798ee8b.... I just realized they disabled it anyway, and seems like it's disabled up until today. 🥲 

Oh well. At least I can remove it then. xD

RazrFalcon:
Oh boy... RIP

LaurenzV:
I have a question. In harfbuzz, there is this one method:

```cpp
static inline bool chain_context_apply_lookup (hb_ot_apply_context_t *c,
					       unsigned int backtrackCount,
					       const HBUINT16 backtrack[],
					       unsigned int inputCount, /* Including the first glyph (not matched) */
					       const HBUINT16 input[], /* Array of input values--start with second glyph */
					       unsigned int lookaheadCount,
					       const HBUINT16 lookahead[],
					       unsigned int lookupCount,
					       const LookupRecord lookupRecord[],
					       ChainContextApplyLookupContext &lookup_context)
{
// Implementation...
}
```

But it seems like in rustbuzz, the same piece of code (with only some variations) appears twice:

```rs
impl Apply for ChainedContextLookup<'_> {
    fn apply(&self, ctx: &mut ApplyContext) -> Option<()> {
      // Implementation
    }
}
```

```rs
fn apply_chain_context(
    ctx: &mut ApplyContext,
    backtrack: LazyArray16<u16>,
    input: LazyArray16<u16>,
    lookahead: LazyArray16<u16>,
    match_funcs: [&MatchFunc; 3],
    lookups: LazyArray16<SequenceLookupRecord>,
) -> Option<()>
```

Do you know why that is? Or am I missing something here?

RazrFalcon:
I guess you can try merging them. I don't think there is a reason why it's done that way.
But again, that's not my code. If you look at git blame you would see. Most of the OpenType code aren't mine.

LaurenzV:
Uff okay, that was tough... So:

1) This PR depends on https://github.com/RazrFalcon/ttf-parser/pull/142, unless you have idea on how we could prevent exposing these fields in `ttf-parser`.

2) The main part of this PR is the addition of `unsafe_to_concat` etc. which required quite a few other refactorings because the code was sometimes very different from `harfbuzz`, but the way it was written didn't allow for implementing those new changes... But now in the end I realized that, if I understood correctly `unsafe_to_concat` etc. API seems to mainly be for the `verify` API, which we don't have yet... 

And since we don't have it yet we also can't pass the flag when testing to make sure it actually works, so it's kind of untested... With that said, I did try really hard to not miss anything, but obviously there still could be bugs. But I think it's not a big deal since we don't offer this API anyway, and all of the old tests (plus the ones that were added) still pass. But just be prepared that once we do implement the verify API, there might be test cases failing. But I think it makes sense to first focus on backporting what we already have, and only then worry about implementing new APIs.

3) I still have some unresolved "yellow" commits above, would be nice if you could take a look!

RazrFalcon:
I really don't like the `ttf-parser` change. The caller must not know about such details. Will see what we can do about it.

LaurenzV:
Sure, if you have ideas let me know. I tried to mirror the `harfbuzz` code as much as possible, which required exposing these. And I think the previous code also was a bit wrong.

src/complex/universal.rs
```diff
@@ -23,6 +23,7 @@ pub const UNIVERSAL_SHAPER: ComplexShaper = ComplexShaper {
 };
 
 pub type Category = u8;
+#[allow(dead_code)]
```
RazrFalcon, line 26: ?

src/ot/matching.rs
```diff
@@ -111,61 +109,71 @@ pub fn match_input(
...
-        positions,
-        total_component_count,
-    })
+    *end_position = iter.index() + 1;
+
+    if let Some(p_total_component_count) = p_total_component_count {
+        *p_total_component_count = total_component_count;
+    }
+
+    return true;
```
RazrFalcon, line 133: Just `true` is enough.

src/ot/matching.rs
```diff
@@ -111,61 +109,71 @@ pub fn match_input(
...
+        let mut unsafe_from = 0;
+        if !iter.prev(Some(&mut unsafe_from)) {
+            *match_start = unsafe_from;
+            return false;
         }
     }
 
-    Some(iter.index())
+    *match_start = iter.index();
+    return true;
```
RazrFalcon, line 154: Just `true` is enough.

src/ot/matching.rs
```diff
@@ -111,61 +109,71 @@ pub fn match_input(
...
+    match_start: &mut usize,
+) -> bool {
     let mut iter = SkippyIter::new(ctx, ctx.buffer.backtrack_len(), backtrack_len, true);
     iter.enable_matching(match_func);
 
     for _ in 0..backtrack_len {
-        if !iter.prev() {
-            return None;
+        let mut unsafe_from = 0;
+        if !iter.prev(Some(&mut unsafe_from)) {
```
RazrFalcon, line 147: Jesus... Can you point how it looks in harfbuzz?

RazrFalcon:
The changes looks fine at the first glance. And there we're a lot of changes... But if tests are passing - we're fine.

>Use invisible-glyph for spaces if font has no ASCII space

Yes, the Rust code is very different, but here it is: https://github.com/RazrFalcon/rustybuzz/blob/7b5c622fcd33e0879b2dd99e4a436cff774af749/src/fallback.rs#L413-L416

But we don't have `x_scale`/`y_scale` anyway, so...

>Add hb_segment_properties_overlay()

I don't think we have `hb_buffer_append`, so ignore.

>[buffer] Clean up internal state bookkeeping

We don't have `Buffer::reset`, but we have `Buffer::clear`.
And it seems we should reset `cluster_level` as well.

>[machinery] Make accelerator lazy-loader call Xinit/Xfini

Completely irrelevant. 

>[unsafe-to-concat] Adjust end conditions

I have no idea... Let's ignore for now.

>[test] Add test-serialize

Irrelevant. 

>[buffer] Add HB_BUFFER_FLAG_VERIFY

Brrrr... Let's ignore it for now. It's an optional feature after all.

RazrFalcon:
As for ttf-parser changes, I will look into them tomorrow. But aren't we decided to ignore `HB_SPLIT_KERN`? 
Can you point where you use new ttf-parser API and corresponding harfbuzz code?

behdad:
> As for ttf-parser changes, I will look into them tomorrow. But aren't we decided to ignore `HB_SPLIT_KERN`?

Yeah. Ignore it.

behdad:
> > [unsafe-to-concat] Adjust end conditions
> 
> I have no idea... Let's ignore for now.

This one port please if you can.

LaurenzV:
> This one port please if you can.

Would you mind explaining it to me? My understanding would be:
**Before**: if end is unintialized (= -1), then set it to len
**After**: always set it to the min of end and len, so if end is uninitialized (= -1), then it stays -1?

But that doesn't sound right...

RazrFalcon:
@LaurenzV  Also, while I know that porting harfbuzz is pain, I'm open to any ideas how we can simplify it. One I have in mind are:

1. Try to make code to look as close as possible to harfbuzz one. The current one is too idiomatic which causes only issues.
2. Try to use the same files structure?
3. Add original names to all the functions to simplify search.

When I was originally porting harfbuzz to Rust I've tried to make the code as idiomatic and as nice as possible, but it was a big mistake... It makes reading code way nicer than original, but hurts porting immensely.

behdad:
> Would you mind explaining it to me? My understanding would be: **Before**: if end is unintialized (= -1), then set it to len **After**: always set it to the min of end and len, so if end is uninitialized (= -1), then it stays -1?
> 
> But that doesn't sound right...

This is `(unsigned) -1`...  The difference is if somehow `end == len + 1`; we want to set end to len or BAD things can happen.

src/complex/universal.rs
LaurenzV: It's because of this commit (already from a port in the past): https://github.com/harfbuzz/harfbuzz/commit/564afc425

It makes those 4 constants become unused, thus resulting in a warning:
```rs
    pub const G: u8 = 49; // HIEROGLYPH
    pub const J: u8 = 50; // HIEROGLYPH_JOINER
    pub const SB: u8 = 51; // HIEROGLYPH_SEGMENT_BEGIN
    pub const SE: u8 = 52; // HIEROGLYPH_SEGMENT_END
```

But since they haven't removed it from harfbuzz either, I just decided to surpress the warnings instead.

src/ot/matching.rs
LaurenzV: Here:

https://github.com/harfbuzz/harfbuzz/blob/ac46c3248e8b0316235943175c4d4a11c24dd4a9/src/hb-ot-layout-gsubgpos.hh#L1205-L1230

That pattern appears a couple of times unfortunately (i.e. having arguments with default values), which doesn't exist in Rust... So in order to stay as close to the original as possible, I decided to use options instead and unwrap them with the default value in the method. Which results in a lot of clutter, but I think that's as close as it gets to the original code.

LaurenzV:
> This is (unsigned) -1... The difference is if somehow end == len + 1; we want to set end to len or BAD things can happen.

Okay, thanks for clarifying! I added it.

LaurenzV:
> Try to make code to look as close as possible to harfbuzz one. The current one is too idiomatic which causes only issues.

Yes, definitely. Idiomatic code is nice, but only if we can stick close to the harfbuzz code, otherwise it will haunt us in the future... One example being the `Matched` struct that I had to dissolve in order to port this time's changes.

> Try to use the same files structure?

In general yes, although for me that wasn't as big of a problem. I think it's more important that the functions are named the same (we don't even need to link the original harfbuzz function, just name it the same works I think). My two main ways of finding the corresponding code locations was 1. search the function names and, believe it or not, what actually helped most was searching for comments that were the same.

Another thing that tripped me up was the way that for example for `PairPos` adjustment, we handle Format1 and Format2 in the same function, while in harfbuzz they are different functions. So when a change happens to only one of them, things get a bit tricky... Or another problem was that for some reason we seem to have two kerning functions? While in harfbuzz I found only one corresponding one. But I think as long as we comment these deviations, it shouldn't be too bad.

LaurenzV:
> Can you point where you use new ttf-parser API and corresponding harfbuzz code?

Sure thing.

# Change 1

```rs
pub fn size(self) -> usize {
        // The high 8 bits are not used, so make sure we ignore them using 0xFF.
        u16::SIZE * self.len()
}

  /// The number of ones in the underlying `ValueFormatFlag`.
  pub fn len(self) -> usize {
      usize::num_from(self.0.count_ones())
  }
```

So here I introduced the `len` method. Before, it was just:
```rs
fn size(self) -> usize {
        // The high 8 bits are not used, so make sure we ignore them using 0xFF.
        u16::SIZE * usize::num_from(self.0.count_ones())
 }
```

Here is the apply method for PosFormat2:

https://github.com/harfbuzz/harfbuzz/blob/ac46c3248e8b0316235943175c4d4a11c24dd4a9/src/hb-ot-layout-gpos-table.hh#L1555-L1666

Note in particular this `len2` variable, whose value is derived from the `get_len` method. I don't know how the previous code worked, but it didn't seem to directly correspond to the harfbuzz code...

That's for PairPos1:

https://github.com/harfbuzz/harfbuzz/blob/ac46c3248e8b0316235943175c4d4a11c24dd4a9/src/hb-ot-layout-gpos-table.hh#L1357-L1374

Although I just noticed they don't have `len2`here... Not sure, maybe I need to look into it again.

# Change 2
The `valueFormatFlags(0)` method we need to add because... well, we can only get `len2` if we can access `valueFormat2`.

# Change 3

The `counts` method I need because of the following piece of code (from PairPos2):

```rs
    if (unlikely (klass1 >= class1Count || klass2 >= class2Count))
    {
      buffer->unsafe_to_concat (buffer->idx, skippy_iter.idx + 1);
      return_trace (false);
    }
```

ttf-parser doesn't know anything about a buffer, so it can't set `unsafe_to_concat`, which means I need to check the condition on the `rustybuzz` side, which means I need access to the counts...

RazrFalcon:
>Yes, definitely. Idiomatic code is nice, but only if we can stick close to the harfbuzz code, otherwise it will haunt us in the future...

I will look into this on weekends. 

>what actually helped most was searching for comments that were the same.

yeah...

> Or another problem was that for some reason we seem to have two kerning functions?

I don't remember, but maybe. harfbuzz relies heavily on C++ templates and it's ofter hard or impossible to port them, because C++ templates are duck typed, unlike Rust one. Which often leads to code duplication.
All I remember is that kern/kerx implementation was an absolute pain and took forever. 

I will see how can we simplify porting. 
The only thing that I do not plan changing is merging `ttf-parser` into `rustybuzz`, which is what `harfbuzz` does. I would prefer to keep them separated.

RazrFalcon:
@LaurenzV As for `ValueRecord`, how about: [positions.patch](https://github.com/RazrFalcon/rustybuzz/files/14297206/positions.patch)

LaurenzV:
Looks sensible, thanks!

LaurenzV:
> The only thing that I do not plan changing is merging ttf-parser into rustybuzz, which is what harfbuzz does. I would prefer to keep them separated.

It does make backporting harder, but understandable that you want to keep them separate, I think it's also useful for the Rust ecosystem to have them separate.

RazrFalcon:
Ready to merge?

As for ttf-parser, while it does make porting a bit harder, I think it's better to have parsing and shaping completely separated. 
harfbuzz has a way too low-level access to TrueType internals. I don't like it. It makes code harder to follow.

LaurenzV:
From my side yes!

RazrFalcon:
Done.

I'm not sure if you still have energy working on it, but I plan to start renaming stuff on weekends, which would make merging PRs impossible. Therefore I would suggest to wait a bit.

LaurenzV:
I'm more or less done with 4.0 (started working on it on a separate branch, just need to rebase), apart from some unclarities. Will open a PR soon. Then I can wait for a while if you want.

PS: Starting from 4.1, they seem to have a couple of commits that reorganize the structure:
image

So maybe you can base it of this then, to save us double the work. But let's see how 4.0 goes.

RazrFalcon:
Got it. Thanks!

khaledhosny:
> 2\. if I understood correctly `unsafe_to_concat` etc. API seems to mainly be for the `verify` API, which we don't have yet...

Verify is a way of testing that these APIs work as intended. `unsafe_to_concat` is for optimizing re-shaping after line breaking. https://github.com/harfbuzz/harfbuzz/issues/1463

LaurenzV:
Thanks for clarifying!

PR #96, opened 2024-02-14
Fix test cmap_1_004 test

LaurenzV:
We just didn't parse the flag... Idk why the comment said we don't use it? But it fixes the test.


LaurenzV:
Turns out it fixes the other test cases, too... However, i will just integrate it into #95 instead.

PR #98, opened 2024-02-16
Sync with 4.1

LaurenzV:
| Status | Commit message                                                         | HB                                                            | Comments                                                                      |
| ------ | ---------------------------------------------------------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------- |
| 🟢     | [USE] Treat visible viramas like dependent vowels                      | [Link](https://github.com/harfbuzz/harfbuzz/commit/49fb8f907) |                                                                               |
| ⚪️     | Make load_num_glyphs_from_loca conditional on HB_NO_BORING_EXPANSION   | [Link](https://github.com/harfbuzz/harfbuzz/commit/965cf1d66) |                                                                               |
| ⚪️     | [baseline] Use ot-metrics fallback API                                 | [Link](https://github.com/harfbuzz/harfbuzz/commit/6bf8f0a38) |                                                                               |
| ⚪️     | [metrics] Simplify x-height fallback                                   | [Link](https://github.com/harfbuzz/harfbuzz/commit/21f5ef56f) |                                                                               |
| ⚪️     | [layout] Whitespace                                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/a8a89b80b) |                                                                               |
| ⚪️     | [baseline] Fix HB_NO_METRICS build                                     | [Link](https://github.com/harfbuzz/harfbuzz/commit/cad2fe8e6) |                                                                               |
| ⚪️     | [subset] bug fix in prune_langsys                                      | [Link](https://github.com/harfbuzz/harfbuzz/commit/7cb002cb5) |                                                                               |
| 🟢     | restores unintended addition in 43be5ba                                | [Link](https://github.com/harfbuzz/harfbuzz/commit/7bdc20ec8) |                                                                               |
| ⚪️     | [coretext] Remove dead code                                            | [Link](https://github.com/harfbuzz/harfbuzz/commit/1ffe637a0) |                                                                               |
| ⚪️     | Add log base 2 versions of constants.                                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/3125f5ae3) |                                                                               |
| ⚪️     | Use bit shifting instead of multiplying and dividing.                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/4ee00f943) |                                                                               |
| ⚪️     | Fix typo.                                                              | [Link](https://github.com/harfbuzz/harfbuzz/commit/7a1e79c3b) |                                                                               |
| ⚪️     | Add option to insert a sorted arrays of values to sets.                | [Link](https://github.com/harfbuzz/harfbuzz/commit/ef588ea97) |                                                                               |
| ⚪️     | Remove null checks.                                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/7d802994b) |                                                                               |
| ⚪️     | Move fn                                                                | [Link](https://github.com/harfbuzz/harfbuzz/commit/1176620ba) |                                                                               |
| ⚪️     | [set] Fix-up previous commits                                          | [Link](https://github.com/harfbuzz/harfbuzz/commit/4289684cd) |                                                                               |
| 🟢     | [indic] Clear syllables before presentation features                   | [Link](https://github.com/harfbuzz/harfbuzz/commit/90f09b1e8) |                                                                               |
| 🟢     | [indic] Test clearing syllables earlier                                | [Link](https://github.com/harfbuzz/harfbuzz/commit/98b485243) |                                                                               |
| ⚪️     | [set] Fix documentation                                                | [Link](https://github.com/harfbuzz/harfbuzz/commit/116cc6923) |                                                                               |
| ⚪️     | [buffer] Fix out-buffer under memory-alloc failure                     | [Link](https://github.com/harfbuzz/harfbuzz/commit/03085132b) |                                                                               |
| ⚪️     | [cmap] In collect_unicodes() of format 12/13                           | [Link](https://github.com/harfbuzz/harfbuzz/commit/f41945e31) |                                                                               |
| 🟢     | [ot-font] Only use vmtx side-bearing if table exists                   | [Link](https://github.com/harfbuzz/harfbuzz/commit/6d0e67dee) | ported and test cases pass, but maybe check if there is a better way to do it |
| ⚪️     | [hmtx] Change default advance for horizontal direction to upem/2 again | [Link](https://github.com/harfbuzz/harfbuzz/commit/df42d28d1) |                                                                               |
| 🟢     | [ot-font] Use ascent+descent for fallback vertical advance             | [Link](https://github.com/harfbuzz/harfbuzz/commit/bf03d7e96) | same                                                                          |
| 🟢     | [ot-font] Vertically center glyph in vertical writing fallback         | [Link](https://github.com/harfbuzz/harfbuzz/commit/1,45E+07)  | same                                                                          |
| 🟢     | Update tests for recent changes                                        | [Link](https://github.com/harfbuzz/harfbuzz/commit/36b8f9741) |                                                                               |
| ⚪️     | Merge pull request #3497 from harfbuzz/vertical-origin                 | [Link](https://github.com/harfbuzz/harfbuzz/commit/280355b31) |                                                                               |
| ⚪️     | [buffer] Whitespace                                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/d35c73cd3) |                                                                               |
| ⚪️     | Remove old TODO file                                                   | [Link](https://github.com/harfbuzz/harfbuzz/commit/0fec8ad84) |                                                                               |
| ⚪️     | 4.1                                                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/c36844d6d) |


LaurenzV:
That one was a bit of a struggle once again due to diverging code bases. :( I tried my best, and the new test cases pass for now. But once you're done with what you want to do I will try to look into whether we can make that part of the codebase closer to what harfbuzz actually does. But for now it should be fine, and you can keep working on the [reorg] commits that are up next, if you feel like doing so.

LaurenzV:
Also https://github.com/RazrFalcon/ttf-parser/pull/143 needs to be merged first, though.

LaurenzV:
Actually, I don't think we have to change to returning a boolean for  `glyph_extents`. We can just return the default instead in the new case. If you want, I can change it back. But if you think it's better to make it look more similar to harfbuzz, I'll leave it the way it is.

RazrFalcon:
Can you point me to the `glyph_extents` change in the original code? Because the `if !glyf_table.has_contours(glyph) {` check is very weird. Why it doesn't check CFF as well?

Otherwise looks good.

RazrFalcon:
Hmm... then your `glyf` check is incorrect. What you check is if the `glyf` table has a data block for the selected glyph. But what harfbuzz checks is if this data block has zero contours. Which is the same as `Face::glyph_bounding_box().is_none()`

LaurenzV:
So should I do this:

```rs
if self.ttfp_face.tables().glyf.is_some() {
    if self.ttfp_face.glyph_bounding_box(glyph).is_none() {
        // Empty glyph; zero extents.
        return true;
    }
}

let Some(bbox) = self.ttfp_face.glyph_bounding_box(glyph) else {
    return false;
};
```

or just:

```rs
let Some(bbox) = self.ttfp_face.glyph_bounding_box(glyph) else {
    return true;
};
```

RazrFalcon:
Hard to say. I'm not really sure what harfbuzz is doing here. A concept of an "empty" glyph is very abstract in TrueType.
To my understanding, harfbuzz returns false only on parsing error. But ttf-parser never returns an error during glyph outlining. From the docs:

>Face::outline_glyph
>Returns `None` when glyph has no outline or on error.

To mimic harfbuzz here we need to heavily update ttf-parser API, which I would rather do later.
So for now, let's do:

```rust
let bbox = self.ttfp_face.glyph_bounding_box(glyph);

if self.ttfp_face.tables().glyf.is_some() && bbox.is_none() {
    // Empty glyph; zero extents.
    return true;
}

let Some(bbox) = bbox else {
    return false;
};
```

Hopefully this would be enough for now.

LaurenzV:
Done.

LaurenzV:
Yeah it's tricky... Maybe we should somehow change ttf-parser to also allow access the low-level stuff, or maybe it should be a separate crate and ttf-parser is then just an abstraction layer on top of it. Not sure... But one step at a time.

RazrFalcon:
Yep, a lot of ideas, not a lot of time.

PR #110, opened 2024-06-17
Backport to 6.0.0

LaurenzV:
Luckily for us, lots of commits that don't apply to us! But a couple of things I wasn't 100% sure about, again:

| Status | Commit message                                                        | HB                                                            |                                                                                                                                                             |
| ------ | --------------------------------------------------------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------- |
| 🟢     | [indic] Order left-matras inside-out                                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/f53ef69d5) |                                                                                                                                                             |
| 🟢     | [GPOS.PairPos] Adjust unsafe-to-break for non-zero ValueFormat2       | [Link](https://github.com/harfbuzz/harfbuzz/commit/ea63e95e4) |                                                                                                                                                             |
| 🟢     | [indic-table] Minor adjust empty lines                                | [Link](https://github.com/harfbuzz/harfbuzz/commit/57de568aa) |                                                                                                                                                             |
| 🟡     | [gsubgpos] Optimize set-digest initialization                         | [Link](https://github.com/harfbuzz/harfbuzz/commit/a053b84cb) | im not sure about these, but as far as I can tell we don't have anything related to digest, right? Is this only necessary for the whole accelerators story? |
| 🟡     | [layout] Use a buffer digest for GPOS to skip whole lookups           | [Link](https://github.com/harfbuzz/harfbuzz/commit/15b6c3259) |                                                                                                                                                             |
| 🟡     | [layout] Use buffer-digest for GSUB as well                           | [Link](https://github.com/harfbuzz/harfbuzz/commit/654a2eafc) |                                                                                                                                                             |
| 🟡     | [buffer] Add .digest() and use                                        | [Link](https://github.com/harfbuzz/harfbuzz/commit/dff1b809a) |                                                                                                                                                             |
| 🟢     | [layout] Only update buffer digest if buffer changed by a pause       | [Link](https://github.com/harfbuzz/harfbuzz/commit/27a8fe7d5) |                                                                                                                                                             |
| 🟢     | [layout] Comment                                                      | [Link](https://github.com/harfbuzz/harfbuzz/commit/658f8f439) |                                                                                                                                                             |
| 🟢     | [thai] Use smaller type for arrays                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/f9d7b303e) |                                                                                                                                                             |
| 🟡     | [gsubgpos] Skippy-iter: Prefer correctness to performance             | [Link](https://github.com/harfbuzz/harfbuzz/commit/7ec1c41a5) | I did translate those, but I'm not sure if the casting is optimal, would like to get some feedback on that                                                  |
| 🟡     | [gsubgpos] Comment                                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/c15efdec4) |                                                                                                                                                             |
| 🟡     | [gsubgpos] No logic-change minor rewrite                              | [Link](https://github.com/harfbuzz/harfbuzz/commit/42681bdb5) |                                                                                                                                                             |
| 🟡     | [gsubgpos] Conditionalize skippy on unsafe-to-concat                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/1fa64c0c2) |                                                                                                                                                             |
| 🟡     | [skippy-iter] Fix two logic errors                                    | [Link](https://github.com/harfbuzz/harfbuzz/commit/0c70bc7f3) |                                                                                                                                                             |
| 🟢     | [PairPos] Another attempt at fixing unsafe-to-break with ValueFormat2 | [Link](https://github.com/harfbuzz/harfbuzz/commit/05aa084e6) |                                                                                                                                                             |
| 🟢     | [indic] Support <U+0A02                                               | [Link](https://github.com/harfbuzz/harfbuzz/commit/260df1fa3) |                                                                                                                                                             |
| 🟢     | COLRv1: use ClipBoxes for extents                                     | [Link](https://github.com/harfbuzz/harfbuzz/commit/7a748ad4a) |                                                                                                                                                             |
| 🟢     | [COLR] Return false from get_extents if table version not 1           | [Link](https://github.com/harfbuzz/harfbuzz/commit/d0ee5a452) |                                                                                                                                                             |
| 🟡     | [COLR] Apply variations in get_extent                                 | [Link](https://github.com/harfbuzz/harfbuzz/commit/9a3f0be29) | from what I can tell harfbuzz always floors in this case? should we do the same or just use the round function as I did?                                    |


src/hb/ot_layout_gsubgpos.rs
```diff
@@ -240,7 +240,20 @@ impl<'a, 'b> skipping_iterator_t<'a, 'b> {
...
+        if self
+            .ctx
+            .buffer
+            .flags
+            .contains(BufferFlags::PRODUCE_UNSAFE_TO_CONCAT)
+        {
+            stop = self.buf_len as i32 - 1;
+        }
+
+        while (self.buf_idx as i32) < stop {
```
LaurenzV, line 256: Those are the weird casts I had to do, but since harfbuzz does the same it should be okay?

src/hb/ot_layout_gsubgpos.rs
```diff
@@ -273,7 +286,19 @@ impl<'a, 'b> skipping_iterator_t<'a, 'b> {
...
+
+        let mut stop = self.num_items - 1;
+
+        if self
+            .ctx
+            .buffer
+            .flags
+            .contains(BufferFlags::PRODUCE_UNSAFE_TO_CONCAT)
+        {
+            stop = 1 - 1;
```
RazrFalcon, line 298: Isn't this just zero? Seems like it was fixed in harfbuzz later. Don't see it in master anymore.

RazrFalcon:
>[gsubgpos] Optimize set-digest initialization | im not sure about these, but as far as I can tell we don't have anything related to digest, right? Is this only necessary for the whole accelerators story?

Yes, we do not have `hb_set_digest_t`. Don't remember why.

>[gsubgpos] Skippy-iter: Prefer correctness to performance | I did translate those, but I'm not sure if the casting is optimal, would like to get some feedback on that

Not gonna lie, as long as tests are passing - I'm fine with it. 

>[COLR] Apply variations in get_extent | from what I can tell harfbuzz always floors in this case? should we do the same or just use the round function as I did?

We cannot floor/ceil/round in ttf-parser because of `no_std`. So as long as it doesn't affects - it's fine.
Afaik we have a couple of "failing" tests already due to different rounding. This is sort of inevitable. We can think of it float precision as an implementation detail.

src/hb/ot_layout_gsubgpos.rs
LaurenzV: It is... But that's how it was in the commit 😆. https://github.com/harfbuzz/harfbuzz/commit/1fa64c0c2

But I can obviously change it.

LaurenzV:
Okay! Then I guess this should be it! 

Looking at the commit history it seems like a lot of future commits are about "irrelevant" stuff (e.g. WASM API and the draw API), so I'm optimistic, but I don't want to jinx it either.

src/hb/ot_layout_gsubgpos.rs
behdad: Yeah, I really wrote a 0 that way. There was an unimportant reason for it. :))

behdad:
> > [gsubgpos] Optimize set-digest initialization | im not sure about these, but as far as I can tell we don't have anything related to digest, right? Is this only necessary for the whole accelerators story?
> 
> Yes, we do not have `hb_set_digest_t`. Don't remember why.

That object is worth porting, since it's a major optimization. It's not too much effort.

LaurenzV:
> That object is worth porting, since it's a major optimization. It's not too much effort.

How about I open an issue for it, so we don't forget about it? The thing is that right now I think the main priority (at least for me) is to get the current code and logic we have synced up with the newest harfbuzz, because what matters most for rustybuzz is correctness. Once we are synced up, we can always work on porting the other optimizations / missing features to rustybuzz, right? 😄  

But I will let @RazrFalcon decide.

behdad:
> > That object is worth porting, since it's a major optimization. It's not too much effort.
> 
> How about I open an issue for it, so we don't forget about it? The thing is that right now I think the main priority (at least for me) is to get the current code and logic we have synced up with the newest harfbuzz, because what matters most for rustybuzz is correctness. Once we are synced up, we can always work on porting the other optimizations / missing features to rustybuzz, right? 😄

Sounds good to me. I might even work on it as a way to find my way around Rust!

LaurenzV:
@behdad I'm currently trying to port to 7.0.0, and there are a couple of `paint-extents` commits which I think are relevant for us (starting from https://github.com/harfbuzz/harfbuzz/commit/268d8b7de). My understand was that by default we use the clip box for the extents, but if no clip box is available we instead "draw" the glyph itself to determine its real bounds, which is then used to determine the actual extents. 

Is my understanding wrong? Because when I tried shaping some text using a font [without a cliplist](https://github.com/googlefonts/color-fonts/blob/main/fonts/test_glyphs-glyf_colr_1_no_cliplist.ttf) I seem to always get extents of zero, even on main:

```
/Users/lstampfl/Programming/GitHub/harfbuzz/buildDir/util/hb-shape /Users/lstampfl/Programming/GitHub/rustybuzz/test_glyphs-glyf_colr_1_no_cliplist.ttf 😀😃😄🤬🤯😳 --no-glyph-names --show-extents
[0=0+600<0,0,0,0>|0=1+600<0,0,0,0>|0=2+600<0,0,0,0>|0=3+600<0,0,0,0>|0=4+600<0,0,0,0>|0=5+600<0,0,0,0>]
```

RazrFalcon:
I do not mind adding `hb_set_digest_t`, as long as someone is willing to implement it.

behdad:
@LaurenzV Did you figure out the COLRv1 clipping thing?

LaurenzV:
Yeah, I just forgot that the test font doesn't contain any actual emojis but only other Unicode codepoints. 😅 It works as expected with the right input.

PR #113, opened 2024-06-22
Run Apple tests in CI

LaurenzV:
Nearly done, just one failing test case I need to investigate.


LaurenzV:
@behdad Do you know whether it's a bug or "intended" that the following two commands yield different results?

```
hb-shape "/System/Library/Fonts/Supplemental/Devanagari Sangam MN.ttc" --unicodes U+092D,U+0941 --no-glyph-names
[160=0+1339|1324=0@-296,11+0]
```

```
hb-shape "/System/Library/Fonts/Supplemental/Devanagari Sangam MN.ttc" --unicodes U+092D,U+0941 --shaper=coretext --no-glyph-names
[160=0+1043|1324=0@0,11+296]
```

I'm not sure if passing `--shaper=coretext` is always supposed to have the same results.

RazrFalcon:
So we did had one bug in `ot_shape.rs`, still way better then I was expecting. But I guess this code haven't changed since I last tested it.

Having our copy of `macos.tests` is meh. Can't we simply add some tests to the ignore list?
And you have already added hash "support" to the Python script.

.github/workflows/main.yml
```diff
@@ -7,11 +7,11 @@ env:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        rust:
-          - stable
+        os: [ubuntu-latest, macos-latest]
```
RazrFalcon, line 13: I think we should hardcode macos version, because macOS 14 to 15 update will definitely break our CI.

LaurenzV: Indeed.

LaurenzV:
> So we did had one bug in ot_shape.rs, still way better then I was expecting. But I guess this code haven't changed since I last tested it.

Two more bugs actually. But yes, luckily not a lot of bugs.

> Having our copy of macos.tests is meh. Can't we simply add some tests to the ignore list?

The thing is that some of the font paths changed apparently, meaning that we would lose many tests just because the font path changed, which would be kind of a waste... I don't think `macos.tests` changes often, so I hope it shouldn't be a big deal. It's not ideal, but certainly better than having no tests at all!

> And you have already added hash "support" to the Python script.

Well, right now we basically ignore them, because as mentioned in the file I don't think we need to bother with running on different MacOS versions, we always run on MacOS 14 for now, and we use the output from harfbuzz as our test oracle. I think that is the cleanest approach for now.

LaurenzV:
Alright, it works now. :D So we can merge this after https://github.com/RazrFalcon/ttf-parser/pull/157 is merged.

RazrFalcon:
Ok, makes sense.

LaurenzV:
Nice! I'll attempt to port the `aat` changes myself next, hopefully it will work.

PR #114, opened 2024-06-24
Sync to 7.3.0

LaurenzV:
A couple of points I wasn't sure about / need to discuss, but other than that nothing crazy, luckily:

| Status | Commit message                                             | HB                                                            |                                                                                                 |
| ------ | ---------------------------------------------------------- | ------------------------------------------------------------- | ----------------------------------------------------------------------------------------------- |
| 🟢     | [MarkBase] Adjust base-finding logic                       | [Link](https://github.com/harfbuzz/harfbuzz/commit/adccc5355) |                                                                                                 |
| 🟢     | [GPOS] Fix indexing in MarkLigPos                          | [Link](https://github.com/harfbuzz/harfbuzz/commit/c67c0086e) |                                                                                                 |
| 🟡     | [GSUB] Support SingleSubst in get_glyph_alternates         | [Link](https://github.com/harfbuzz/harfbuzz/commit/7327006d6) | I guess not relevant? seems more like an API thing, not actually relevant for shaping           |
| 🟢     | Add Tifinagh to list of both-directions scripts            | [Link](https://github.com/harfbuzz/harfbuzz/commit/2d8634624) |                                                                                                 |
| 🟢     | [indic] Tighten up base-finding                            | [Link](https://github.com/harfbuzz/harfbuzz/commit/453ded053) |                                                                                                 |
| 🟢     | [syllabic] Actually clear syllables                        | [Link](https://github.com/harfbuzz/harfbuzz/commit/c5f3b3feb) |                                                                                                 |
| 🟡     | [syllabic] Better fix for previous issue                   | [Link](https://github.com/harfbuzz/harfbuzz/commit/63afb4f2e) | unsure about this one, our skippy iter only has syllable but not a bool "per_syllable">         |
| 🟢     | [match_input] Micro-optimize                               | [Link](https://github.com/harfbuzz/harfbuzz/commit/49ac5e11e) |                                                                                                 |
| 🟢     | [PairPos2] Micro-optimize                                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/599671543) |                                                                                                 |
| 🟡     | [PairPosFormat2] Micro-optimize and don't kern if class2=0 | [Link](https://github.com/harfbuzz/harfbuzz/commit/95f155573) | not sure I can really port this... I remember we ported this in the past and it was a huge pain because ttf-parser doesn't give that much low-level access |
| 🟡     | [Ligature] Speed up                                        | [Link](https://github.com/harfbuzz/harfbuzz/commit/7881eadff) | do we need to port this since it's behind a compilation flag? probably easier to ignore it      |
| 🟡     | [Ligature] Minor tweak to recent code                      | [Link](https://github.com/harfbuzz/harfbuzz/commit/51061d285) |                                                                                                 |
| 🟡     | [Ligature] Micro-optimize                                  | [Link](https://github.com/harfbuzz/harfbuzz/commit/0fe90ebc0) |                                                                                                 |
| 🟡     | [Ligature] Micro-optimize more                             | [Link](https://github.com/harfbuzz/harfbuzz/commit/ddd6c2e7a) | same as above                                                                                   |
| 🟡     | [Ligature] Use slow path if 2 or fewer ligatures           | [Link](https://github.com/harfbuzz/harfbuzz/commit/ae0fe02d1) |                                                                                                 |
| 🟡     | [GPOS] Optimize iterator reset                             | [Link](https://github.com/harfbuzz/harfbuzz/commit/fb795dc3c) | we dont have the whole iter thing with init and reset right?                                    |
| 🟡     | [layout] Don't init iters successively multiple times      | [Link](https://github.com/harfbuzz/harfbuzz/commit/3f2401e2f) | same                                                                                            |

### DIFF
--- a/src/hb/common.rs
+++ b/src/hb/common.rs
@@ -144,7 +144,8 @@ impl Direction {
             // https://github.com/harfbuzz/harfbuzz/issues/1000
             script::OLD_HUNGARIAN |
             script::OLD_ITALIC |
-            script::RUNIC => {
+            script::RUNIC |
+            script::TIFINAGH => {
                 None
             }
 
--- a/src/hb/ot/layout/GPOS/mark_base_pos.rs
+++ b/src/hb/ot/layout/GPOS/mark_base_pos.rs
@@ -31,7 +31,8 @@ impl Apply for MarkToBaseAdjustment<'_> {
         while j > ctx.last_base_until as usize {
             let mut _match = iter.match_(&buffer.info[j - 1]);
             if _match == match_t::MATCH {
-                if !accept(buffer, j - 1) {
+                // https://github.com/harfbuzz/harfbuzz/issues/4124
+                if !accept(buffer, j - 1) && !self.base_coverage.contains(buffer.info[j - 1].as_glyph()) {
                     _match = match_t::SKIP;
                 }
             }
--- a/src/hb/ot/layout/GPOS/mark_base_pos.rs
+++ b/src/hb/ot/layout/GPOS/mark_base_pos.rs
@@ -32,7 +32,9 @@ impl Apply for MarkToBaseAdjustment<'_> {
             let mut _match = iter.match_(&buffer.info[j - 1]);
             if _match == match_t::MATCH {
                 // https://github.com/harfbuzz/harfbuzz/issues/4124
-                if !accept(buffer, j - 1) && !self.base_coverage.contains(buffer.info[j - 1].as_glyph()) {
+                if !accept(buffer, j - 1)
+                    && !self.base_coverage.contains(buffer.info[j - 1].as_glyph())
+                {
                     _match = match_t::SKIP;
                 }
             }
--- a/src/hb/ot/layout/GPOS/mark_lig_pos.rs
+++ b/src/hb/ot/layout/GPOS/mark_lig_pos.rs
@@ -26,8 +26,6 @@ impl Apply for MarkToLigatureAdjustment<'_> {
         while j > ctx.last_base_until as usize {
             let mut _match = iter.match_(&buffer.info[j - 1]);
             if _match == match_t::MATCH {
-                // Due to borrow checker, we cannot assign directly to ctx.last_base in the
-                // for loop, unlike in harfbuzz.
                 ctx.last_base = j as i32 - 1;
                 break;
             }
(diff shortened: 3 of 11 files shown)


RazrFalcon:
>[GSUB] Support SingleSubst in get_glyph_alternates	Link	I guess not relevant? seems more like an API thing, not actually relevant for shaping

Yep.

>[syllabic] Better fix for previous issue	Link	unsure about this one, our skippy iter only has syllable but not a bool "per_syllable">

Looks like it was added in v4: https://github.com/harfbuzz/harfbuzz/commit/044d7a06db552e1564b8575f4d23798f009d9dde
We should add it too. Strange that it works at all without it...

>[PairPosFormat2] Micro-optimize and don't kern if class2=0	Link	not sure I can really port this... I remember we ported this in the past and it was a huge pain because ttf-parser doesn't give that much low-level access

Let's ignore.

>[Ligature] Speed up	Link	do we need to port this since it's behind a compilation flag? probably easier to ignore it

Looks like it was changed quite a bit in the master. Nice to have, but can leave it for later.
Can leave a commet/todo in Rust code explaining why it is different. 

>[GPOS] Optimize iterator reset	Link	we dont have the whole iter thing with init and reset right?

No idea... Seems like we never reset a skipping iterator.

LaurenzV:
Hm, yeah there is definitely something fishy... So this `per_syllable` thing is only used in the `set_syllable` function, which in turn is only used in the `reset` function of the iterator... which for some reason, we don't have, along with `init`. I searched for its usages, and there are indeed differences in how we interact with the iterator throughout the code. Not sure why it is that way. I guess for now I'll just add a TODO.

LaurenzV:
There are some more changes regarding how skippy iterators are used, so I'll try to align it in the next PR.

RazrFalcon:
Got it. I remember it was problematic to implement skipping iterator. Probably because of borrow checker. So no surprises it is different.
But `per_syllable` was still added after my initial port.

LaurenzV:
I can see why now... the context itself contains a skipping iterator `iter_input`, and the skipping iterator itself contains a reference to the same context, so it's a circular reference... Yeah this seems pretty hard to deal with. :/

RazrFalcon:
We should start sending patches to harfbuzz 👀

Issue #117, opened 2024-06-26
Embed experimental `wasm` shaper.


asibahi:
Hello.

First of all congratulations on reaching the 8.0.0 milestone. Truly impressive work by @LaurenzV. There is an issue on `harfbuzz` to bump version to 9 :)

I am wondering if it would be within the scope of this project to include API and parsing for the experimental `wasm` shaper in harfbuzz. As far as I can tell, it is "just" an extra font table and some C API. The `wasm` modules are embedded in the font itself. 

Here is the design document, with some fairly interesting examples: https://github.com/harfbuzz/harfbuzz/blob/main/docs/wasm-shaper.md 
This is an insane project built on the shaper: https://github.com/fuglede/llama.ttf

This would allow more people (font designers for example) to experiment with the feature without the massive gate of requiring knowledge on how to build `harfbuzz`, since its not enabled by default and the only program that comes with it enabled is (an experimental version of) FontGoogles.

The main advantage for including this in `rustybuzz` as a non-default feature is that building a rust crate is infinitely simpler than building a C++ library, even I can do it!!

RazrFalcon:
I have no plans on implementing any non-standard TrueType features. But if someone will send a PR - I will merge it.

PR #122, opened 2024-07-03
Draft implementation of wasm shaper

asibahi:
Hello.

working on [this](https://github.com/RazrFalcon/rustybuzz/issues/117)

So I started a basic structure of the `wasm` runner and how to export functions *to* the `wasm` module. I am using the `wasmtime` crate.

The general workflow is this : 
1. look for `Wasm` table. If it is there : run it.
2. the `shape_with` exported function should just always direct to `shape_with_plan`
3. if `Wasm` table is not there or is malformed , use `shape_with_plan`. instead like normal.

I already hit a couple of things I do not understand:

1. the API has a concept of scale, which rustybuzz (afaik) does not have. Some of the harfbuzz examples depend on scale. What would be the correct strategy to deal with this?
2. I have absolutely no idea how to deal with pointers. Using `wasmtime` APIs to "link" functions with pointers is giving me compiler errors. You can see what the compiler points are by uncommenting the marked lines. Something about traits?

This is a draft PR hoping for some feedback and help.

Final edit I swear:
The two places where I am looking at the functions and how they're used are :
1. [The Design document](https://github.com/harfbuzz/harfbuzz/blob/main/docs/wasm-shaper.md)
2. [The `harfbuzz-wasm` crate.](https://github.com/harfbuzz/harfbuzz/blob/main/src/wasm/rust/harfbuzz-wasm/src/lib.rs) (yes that's the entire crate.)

The C++ implementation of the `wasm` api is [here](https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-wasm-api.h) But C++ is black magic to me.


Cargo.toml
```diff
@@ -21,6 +21,7 @@ unicode-bidi-mirroring = "0.3.0"
 unicode-ccc = "0.3.0"
 unicode-properties = { version = "0.1.0", default-features = false, features = ["general-category"] }
 unicode-script = "0.5.2"
+wasmtime = "22.0.0"
```
notgull, line 24: This should at least be optional; I don't want to pull in the entirety of wasmtime and cranelift for a text shaper.

If it were me I would just implement a WASM interpreter inline in rustybuzz. It's not like WASM is that complicated to parse, and it's not like we need WASI IIRC.

notgull (changes requested):

Cargo.toml
asibahi: Yes of course. I would add all the feature flags and such once I am done

LaurenzV:
> the API has a concept of scale, which rustybuzz (afaik) does not have. Some of the harfbuzz examples depend on scale. What would be the correct strategy to deal with this?

How exactly do they depend on it? We indeed don't have scale, so no scale should be applied (or basically a "scale" of 1).

behdad:
> > the API has a concept of scale, which rustybuzz (afaik) does not have. Some of the harfbuzz examples depend on scale. What would be the correct strategy to deal with this?
> 
> How exactly do they depend on it? We indeed don't have scale, so no scale should be applied (or basically a "scale" of 1).

Or rather, scale of UPEM in HB terms.

behdad:
This is nice, particularly because wasmtime seems to have a "fuel" concept to limit CPU usage:

  https://github.com/bytecodealliance/wasmtime/blob/main/examples/fuel.rs

It generally looks like a better runtime than wasm-micro-runtime, which has several problems as a library.

asibahi:
> > > the API has a concept of scale, which rustybuzz (afaik) does not have. Some of the harfbuzz examples depend on scale. What would be the correct strategy to deal with this?
> > 
> > 
> > How exactly do they depend on it? We indeed don't have scale, so no scale should be applied (or basically a "scale" of 1).
> 
> Or rather, scale of UPEM in HB terms.

So I should just return the upem and be done with it?

behdad:
> > > > the API has a concept of scale, which rustybuzz (afaik) does not have. Some of the harfbuzz examples depend on scale. What would be the correct strategy to deal with this?
> > > 
> > > 
> > > How exactly do they depend on it? We indeed don't have scale, so no scale should be applied (or basically a "scale" of 1).
> > 
> > 
> > Or rather, scale of UPEM in HB terms.
> 
> So I should just return the upem and be done with it? This leaves the issue of how to deal with pointers :D

Yes. UPEM works I think. No idea re pointers I'm afraid.

Cargo.toml
notgull: What I'm saying is that wasmtime is definitely overkill for this use case. I doubt that embedded fonts will need WASI or JIT compilation.

asibahi:
Gave more flesh to a few functions. Left plenty of `todo!`s tho. I'd appreciate someone looking at them.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+    // fn font_get_face(font: u32) -> u32;
+    // Creates a new face token from the given font token.
+    let font_get_face = |caller: Caller<'_, ShapingData>, _: u32| {
+        // From HarfBuzz docs:
+        // (In the following functions, a font is a specific instantiation of a face at a
+        // particular scale factor and variation position.)
+        //
+        // I am unsure how to represent that in rustybuzz.
+        // er .. do what here?
+        0
```
RazrFalcon, line 73: What harfbuzz does in this case?

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+    // Just return the upem as rustybuzz has no scale.
+    let font_get_scale =
+        |mut caller: Caller<'_, ShapingData>, _: u32, x_scale: u32, y_scale: u32| {
+            let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+            let upem = caller.data().font.units_per_em();
+
+            memory.write(
+                &mut caller.as_context_mut(),
+                x_scale as usize,
+                &upem.to_le_bytes(), // le or be?
```
RazrFalcon, line 111: Most certainly LE.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+            let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+
+            let mut glyph_extents = hb_glyph_extents_t::default();
+            let ret = caller
+                .data()
+                .font
+                .glyph_extents(GlyphId(glyph as u16), &mut glyph_extents);
+            if ret {
+                // WASM is little endian. MacOS is little endian, I think.
+                // Maybe should use bytemuck as a dependency instead.
```
RazrFalcon, line 136: We already use `bytemuck`

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+            let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+
+            let name = caller
+                .data()
+                .font
+                .glyph_name(GlyphId(glyph as u16))
+                .map(ToOwned::to_owned)
+                .unwrap_or(format!("g{glyph:4}"));
+
+            let name = CString::new(name).unwrap();
```
RazrFalcon, line 165: Are you sure it is C string and not just `[i8; N]`?

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+        |mut caller: Caller<'_, ShapingData>, _: u32, glyph: u32, outline: u32| -> u32 {
+            let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+
+            let builder = CGlyphOutline {
+                n_points: todo!(),
+                points: todo!(),
+                n_contours: todo!(),
+                contours: todo!(),
+            };
+            // also no clue what to do here
```
RazrFalcon, line 209: Hm... We never preserve contours. `ttf-parser` can only return points. So you can try setting `n_contours` to zero. No idea why harfbuzz needs it.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+            let tag = tag.to_be_bytes(); // biggest byte first right ?
+            let Some(table) = caller.data().font.raw_face().table(Tag::from_bytes(&tag)) else {
+                return 0;
+            };
+
+            let length = table.len() as u32;
+
+            let my_blob = Blob {
+                length,
+                data: table.as_ptr() as usize, // is this correct?
```
RazrFalcon, line 242: Hard to say. Ideally it should be `*mut u8`

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,421 @@
...
+
+        std::eprintln!("{msg}"); // maybe?
+    };
+    linker
+        .func_wrap(module_name, "debugprint", debugprint)
+        .ok()?;
+
+    // fn shape_with(font: u32, buffer: u32, features: u32, num_features: u32, shaper: *const u8) -> i32;
+    // Run another shaping engine's shaping process on the given font and buffer. The only shaping engine guaranteed to be available is ot, the OpenType shaper, but others may also be available. This allows the WASM author to process a buffer "normally", before further manipulating it.
+    // I think we should just use the default rustybuzz shaper for now.
```
RazrFalcon, line 345: This means calling `rustybuzz::shape` since we implement just a single shaper.

src/hb/shape_wasm.rs
asibahi: C++ code is inscrutable to me so I don't know. From the hafbuzz-wasm crate it seems to be just a token being passed to the host (not even a pointer).

src/hb/shape_wasm.rs
asibahi: `harfbuzz-wasm` assumes it is a `Cstring` : https://github.com/harfbuzz/harfbuzz/blob/e079dd20304bacd8d753eee2ceacbaa1a3978a34/src/wasm/rust/harfbuzz-wasm/src/lib.rs#L459

src/hb/shape_wasm.rs
asibahi: In my local machine I changed this a bit to picking an arbitrary place in the wasm module's memory (specifically `blob + std::mem::size_of::<Blob>() as u32` and passed that in. My understanding is that pointers *within* the wasm module point essentially to indices to the giant `&[u8]`, and passing host pointers is meaningless.

We will know more once I manage to get something running.

src/hb/shape_wasm.rs
asibahi: Hah right. I need to read up on the API.

src/hb/shape_wasm.rs
RazrFalcon: Good. Just wanted to make sure. Then we can skip length I guess.

src/hb/shape_wasm.rs
asibahi: [that's the header file](https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-wasm-api.h). 

I *think* [that's the function in question](https://github.com/harfbuzz/harfbuzz/blob/e079dd20304bacd8d753eee2ceacbaa1a3978a34/src/hb-wasm-api-font.hh#L47).

src/hb/shape_wasm.rs
asibahi: I think I figured out the "proper" way to handle this.

src/hb/shape_wasm.rs
RazrFalcon: Hm... We do not have a concept of a Font in rustybuzz/ttf-parser. Everything is a Face.
But it returns an index of some sort and not a "face". No idea what this index is suppose to represent.
Better as harfbuzz authors about it.

behdad: If there's no Font in RB, you can just pass the Face as the Font...

The index is what represents the object inside the interpretter. You need a mapper for mapping back and force between pointers and the indices. In HB we use the OBJ2REF and REF2OBJ from our wasm runtime.

asibahi:
Hey

Positive things:
1. (I think) I have a better understanding of how the wasm modules treats memory and pointers. Long story short, `memory` is a giant array of `u8`. And pointers allocated from the module are indices on that array.
2. The file is easier to navigate, generally. The imported functions are now full fledged functions instead of closures written inline.
3. I replaced most of `transmute` with `bytemuck`, as much as I can figure it out.
4. I added a test for the [Calculator example](https://github.com/harfbuzz/harfbuzz-wasm-examples/tree/main/calculator). It doesn't really work tho. See below.

So I am trying right now to make the Calculator example work. It only imports 5 functions of the set, and exports only the needed function `shape`.

As copied from the `wat` file: 
```wasm
(import "env" "buffer_copy_contents" (func (;0;) (type 0)))
(import "env" "buffer_set_contents" (func (;1;) (type 0)))
(import "env" "font_get_glyph" (func (;2;) (type 3)))
(import "env" "font_get_glyph_h_advance" (func (;3;) (type 0)))
(import "env" "debugprint" (func (;4;) (type 2)))
```

The last three functions are straightforward I guess and all the complexity is in getting the first two working. I added a test and everything. The good news is that the function isn't returning or crashing. The bad news is that it is not .. working. I can't tell right now where the issue is, but it is giving me the same input as output. And I am quite sure where the issue is. 

I am thinking this is perhaps due to a misunderstanding, on my part, on how the `hb_buffer_t` works. My current assumption is that, after initializing `UnicodeBuffer`, I take the `info` and `pos` vectors out and pass them into the module. (that's what I am trying to do in `buffer_copy_contents`.) Then after everything, take the `info` and `pos` arrays produced by the vector and put them back into `hb_buffer_t`  before wrapping it into `GlyphBuffer`. (That's what I am trying to do in `buffer_set_contents`. )

Reading the impls of `hb_buffer_t`, there seems to be some assumptions I am not sure I get. Are `info` and `pos` always the same length? (They should be, right?) Why does the buffer return `pos` as `info` sometimes?

When I change the `clear()` functions to the commented out code in line 396, changing nothing else, the output *does* change. Now I get the same number of glyphs I am getting, but all the data is zeroed out. When I tried replacing the `transmutes` with some `bytemuck`, the test crashed somewhere in `buffer.rs`.

PS. Sorry for the messy commit history. It is kind of a reflection of the thought process :smile:

behdad:
> Reading the impls of `hb_buffer_t`, there seems to be some assumptions I am not sure I get. Are `info` and `pos` always the same length? (They should be, right?)

Correct.

> Why does the buffer return `pos` as `info` sometimes?

That's an internal optimization, that the pos array is used as a secondary `info` during the substitution process before positions are used. It's not relevant to the wasm from what I remember.

asibahi:
Oh okay nvm the previous comment. I found a bug in `font_get_glyph` where I don't check if `uvs` equals 0 and convert it to a `char` anyway, and `ttf-paerse`'s `glyph_variation_index` returns `None` when `uvs` is `\0`.

tl;dr : Calculator works now.

RazrFalcon:
Congrats!

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+    _font: u32,
+    codepoint: u32,
+    uvs: u32,
+) -> u32 {
+    let Some(codepoint) = char::from_u32(codepoint) else {
+        return 0;
+    };
+
+    if uvs == 0 {
+        caller.data().font.glyph_index(codepoint).unwrap().0 as u32
```
laurmaedje, line 143: should we also return 0 here instead of unwrapping?

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+
+// fn font_get_glyph_extents(font: u32, glyph: u32, extents: *mut CGlyphExtents) -> bool;
+// Returns the glyph's extents for the given glyph ID at current scale and variation settings.
+fn font_get_glyph_extents(
+    mut caller: Caller<'_, ShapingData>,
+    _font: u32,
+    glyph: u32,
+    extents: u32,
+) -> u32 {
+    let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
```
laurmaedje, line 184: best check upfront (during initialization of the wasm) that the export exists (like [here](https://github.com/typst/typst/blob/2df138a507860dc0cb610943ad6e608a33d03d7d/crates/typst/src/foundations/plugin.rs#L198-L204)) to prevent a panic for a bad font

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+    let mut glyph_extents = hb_glyph_extents_t::default();
+    let ret = caller
+        .data()
+        .font
+        .glyph_extents(GlyphId(glyph as u16), &mut glyph_extents);
+    if ret {
+        memory.write(
+            caller.as_context_mut(),
+            extents as usize,
+            bytemuck::bytes_of(&glyph_extents),
```
laurmaedje, line 195: won't this break on big endian machines?

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+
+    let name = caller
+        .data()
+        .font
+        .glyph_name(GlyphId(glyph as u16))
+        .map(ToOwned::to_owned)
+        .unwrap_or(format!("g{glyph:4}"));
+
+    let name = CString::new(name).unwrap();
+    let name = &name.as_bytes()[..len as usize]; // are names ever non ascii?
```
laurmaedje, line 223: Two things here:
- this will panic with index out of bounds if the `len` exceeds the actual length of the name. 
- if the `len` is less than `name.len() - 1`, it will not copy the null terminator

What you could do is `into_bytes` the name `String`, `truncate` it and then add the null-terminator manually. 

(Could also skip allocating at all when there's a name, but might be unnecessary optimization.)

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+        n_contours: todo!(),
+        contours: todo!(),
+    };
+    // also no clue what to do here
+    // let my_ol = caller.data().outline_glyph(GlyphId(glyph as u16), builder); // ??
+
+    let Ok(()) = memory.write(caller, outline as usize, todo!("builder result goes here")) else {
+        return 0;
+    };
+
```
laurmaedje, line 262: It's indeed interesting. I don't think the `CGlyphOutline` struct makes much sense to have in this file since the struct is meant to be operated on by the wasm module, meaning the pointers in it are wasm-sized, but this struct will have host-sized pointers. Looking at harfbuzz-wasm, the `outline` parameter represents a reference to an already existing, uninitialized `CGlyphOutline` struct in the wasm memory. But it is unclear to me how the `points` and `contours` pointers are supposed to be filled since there is no way to allocate memory in the module and we can't just randomly write somewhere.

edit: same problem as in `face_copy_table`

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+    1
+}
+
+// fn face_copy_table(font: u32, tag: u32, blob: *mut Blob) -> bool;
+// Copies the binary data in the OpenType table referenced by tag into the supplied blob structure.
+fn face_copy_table(mut caller: Caller<'_, ShapingData>, _font: u32, tag: u32, blob: u32) -> u32 {
+    // So here to copy stuff INTO the module, I need to copy it into its heap
+    // I should not assume that there is an area that's not written to,
+    // so the most straightforward way to get "clean" memory is to grow it by one page,
+    // and allocate there.
```
laurmaedje, line 272: yeah, it's very confusing how this is supposed to work.

maybe @behdad can shed some light on this? I see that harfbuzz uses some `wasm_runtime_addr_app_to_native` construct from `wasm-micro-runtime`, but I don't understand how it works conceptually.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+
+    let rb_buffer = &store_data.buffer;
+    let length = rb_buffer.len;
+
+    let pos_loc = eom + length * core::mem::size_of::<hb_glyph_info_t>();
+    let end_loc = pos_loc + length * core::mem::size_of::<GlyphPosition>();
+
+    // This _should_ work ..
+    mem_data[eom..pos_loc].copy_from_slice(bytemuck::cast_slice(&rb_buffer.info));
+    mem_data[pos_loc..end_loc].copy_from_slice(bytemuck::cast_slice(&rb_buffer.pos));
```
laurmaedje, line 326: I think this could panic if the memory didn't grow enough

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+}
+
+// fn debugprint(s: *const u8);
+// Produces a debugging message in the host shaper's log output; the variants debugprint1 ... debugprint4 suffix the message with a comma-separated list of the integer arguments.
+// rust varargs when
+fn debugprint(mut caller: Caller<'_, ShapingData>, s: u32) {
+    let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+
+    // seems reasonable?
+    let bytes = &memory.data(&caller)[s as usize..];
```
laurmaedje, line 387: this also use checked indexing as to not crash anything

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,505 @@
...
+    let msg = CStr::from_bytes_until_nul(bytes)
+        .unwrap_or_default()
+        .to_string_lossy();
+
+    std::eprintln!("{msg}"); // maybe?
```
laurmaedje, line 392: I know that it's all WIP, but if it lands ideally it should log somewhere else

laurmaedje (commented):
@asibahi asked me to give this a quick look, so here are some comments

src/hb/shape_wasm.rs
asibahi: Yes that was me debugging the function. Already fixed locally.

src/hb/shape_wasm.rs
asibahi: Done

src/hb/shape_wasm.rs
asibahi: The `memory.grow(1)` dance later in the file was suggested by helpful souls on the Rust Discord. I havent done it here because I was focused on getting `Calculator-Regular.ttf` working and it doesn't make use of this.

It seems the idiomatic WASM solution for this is to use `ExternRef`, but Rust compiled to `wasm` doesn't naturall support it (tho there are [crates](https://docs.rs/externref/latest/externref/)), *and* the `harfbuzz wasm examples don't make use of it. Maybe the WASM API could be changed but that's a talk for another time.

My issue for *this* function is like @RazrFalcon mentioned that `rustybuzz` doesn't keep the contours, so we might just pass the points. I'd have to test this with the Nastaliq and Ruqaa examples.

src/hb/shape_wasm.rs
asibahi: [This is the function from `wasm-micro-runtime](https://github.com/bytecodealliance/wasm-micro-runtime/blob/0feae05379feb856f6b1af6c0dc5cf57f87155b7/core/iwasm/common/wasm_memory.c#L448).

src/hb/shape_wasm.rs
asibahi: As far as I understand the spec, it grows by one "page", which is 65,536 bytes. That fits 1638 characters (both info and pos). Maybe I can add a check.

laurmaedje: Aren't there multiple infos and pos structs? So if the buffer grows larger it could exceed the page size?

src/hb/shape_wasm.rs
laurmaedje: Okay, so that function simply translates the pointer by the base offset. The interesting one that is used to actually write is `wasm_runtime_module_dup_data`, which ends up calling [`wasm_module_malloc_internal`](https://github.com/bytecodealliance/wasm-micro-runtime/blob/0feae05379feb856f6b1af6c0dc5cf57f87155b7/core/iwasm/interpreter/wasm_runtime.c#L3520), which seems to, depending on the circumstances, use various strategies to allocate some memory it can use.

src/hb/shape_wasm.rs
asibahi: GlyphPosition is size 20. I just divided 65,538 by 40 (both info and pos) to get that number. 1600 characters/letters is like 200 words ish. I don't know if shaper input is usually this large or not.

src/hb/shape_wasm.rs
asibahi: Done

src/hb/shape_wasm.rs
asibahi: I don't think there is a specific logging mechanism in rustybuzz. What do you think, @RazrFalcon ?

src/hb/shape_wasm.rs
asibahi: I don't think that's needed here. as `s` is gived to us by the module and is (probably?) guarabteed to be a valid index.

(Also this worked fine for the Calculator wasm, and printed it and everything.)

src/hb/shape_wasm.rs
asibahi: Looking at the code in `harfbuzz-wasm`, [here](https://github.com/harfbuzz/harfbuzz/blob/e079dd20304bacd8d753eee2ceacbaa1a3978a34/src/wasm/rust/harfbuzz-wasm/src/lib.rs#L154), `n_contours` seems to be necessary, as it is used to draw the curves from which collision calculations are made.

From what I understand from the code, contours here are the indices of which points are `end_pts_of_contours`. Is there a straightforward way to get this info from `ttf-parser`?

src/hb/shape_wasm.rs
asibahi: No idea.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,604 @@
...
+    fn curve_to(&mut self, x1: f32, y1: f32, x2: f32, y2: f32, x: f32, y: f32) {
+        // is this correct ?
+        self.points.push(CGlyphOutlinePoint {
+            x: x1, // same assumption here.
+            y: y1,
+            pointtype: PointType::CubicTo,
+        })
+    }
+
+    fn close(&mut self) {
```
asibahi, line 524: Logically speaking, should this be `len` or `len - 1` ?

src/hb/shape_wasm.rs
laurmaedje: I'm thinking of the case of a font that wants to crash the shaper.

asibahi:
I tried to implement the `outline` function as much as I can glean from the code using it, and a simple impl of `shape_with` Also added a failing test for `ArefRuqaa-Was.ttf`. 

The failure is due to a failure *within* the module, reaching an unreachable state. The same font works in the Wasm version of FontGoggles. Printing the error from my side gives me this :

```
Err(error while executing at wasm backtrace:
    0: 0x16370 - <unknown>!<wasm function 116>
    1: 0x16a7d - <unknown>!<wasm function 127>
    2: 0x17891 - <unknown>!<wasm function 170>
    3: 0x173b2 - <unknown>!<wasm function 150>
    4: 0x17764 - <unknown>!<wasm function 164>
    5: 0x16e96 - <unknown>!<wasm function 136>
    6: 0xeb22 - <unknown>!<wasm function 40>
    7: 0x13870 - <unknown>!<wasm function 77>
    8: 0x111cf - <unknown>!<wasm function 55>
    9: 0x505c - <unknown>!<wasm function 13>

Caused by:
    wasm trap: wasm `unreachable` instruction executed)
```

I know function 13 (the last one there) is the exported `shape` function. No idea beyond that on what's going on.

asibahi:
Managed to get it to work by calling `crate::shape` instead of `crate::shape_with_plan`. 

Two weird issues:
1. The `ArefRuqaa-Wasm` module is breaking (I assume panicking at index bounds ...) when the last letter of the sentence is `.fina`. So when I type `أفشوا السلام بينكم` as input, it fails. When I type `أفشوا السلام بينكم.` (note the period), it runs to completion.
2. Output is not identical to Wasm FontGoggles. Clusters are not identical, and some x_advances are not. I do not know if this is due to different versions of `harfbuzz` or some bug else where .. I have no idea.

src/hb/shape_wasm.rs
asibahi: definitely `len`

src/hb/shape_wasm.rs
RazrFalcon: We do not support big endian machines to begin with.

src/hb/shape_wasm.rs
RazrFalcon: As said before, `ttf-parser` doesn't preserve this info. It would be really hard to do.

Also, `contours` is a `glyf` specific thing. CFF doesn't have them. No idea why WASM table relies on them to begin with.

src/hb/shape_wasm.rs
RazrFalcon: It will not work with no_std builds anyway. So it's easier to simply ignore this function. We can comment it out, in case someone would like to debug it.

RazrFalcon:
>also Quads and Cubes need more than one point pushed into points. 

Quad is 2 points, Cube is 3 points. All of which you can get from OutlineBuilder.

>Not sure how the point type maps to on-curve and off-curve but here we are.

This is a `glyf`-specific thing. CFF doesn't have it.

In general, it seems like `CGlyphOutline` and the whole Wasm table is extremely harfbuzz-specific, meaning a regular text shaper will not be able to support it. @behdad is this expected?

src/hb/shape_wasm.rs
laurmaedje: Is that mentioned somewhere? And are there existing code paths that don't work on big endian?

src/hb/shape_wasm.rs
asibahi: Actually the runtime Typst uses, `wasmi` , can compiled to `no_std`. I’ve been meaning to compile a list of pros and cons for it and `wasmtime` when I ironed out most things.

RazrFalcon: Got it. But there are no print in no_std anyway. And we do no have logging for now.

src/hb/shape_wasm.rs
RazrFalcon: I don't think it's mentioned. All I know is that we do not test for big endian architectures. Mainly because they are so rare.
It might work, might not.

asibahi:
I think I got this as far as I can, tbh. 

The Ruqaa test (with both inputs, see comments) is failing for reasons I cannot understand.. I will try and get an image out of the working output, as the failure there doesn't seem to me to be because of the wasm runtime itself. But I don't want to add `image` and `ab_glyph` as `dev-dependencies`, so the image thing will be out of crate.

RazrFalcon:
A smaller wasm dependency is for sure preferable.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,728 @@
...
+    length: u32,
+    info: u32,
+    position: u32,
+}
+
+unsafe impl bytemuck::Zeroable for CBufferContents {}
+unsafe impl bytemuck::Pod for CBufferContents {}
+
+#[cfg(test)]
+mod wasm_tests {
```
RazrFalcon, line 582: Can you move tests to the tests directory?

Cargo.toml
```diff
@@ -37,6 +38,10 @@ features = [
 default = ["std"]
 std = ["ttf-parser/std"]
 
+# experimental features live here.
+experimental = ["wasm_shaper"]
```
RazrFalcon, line 42: I do not think we need the `experimental` feature. `wasm_shaper` is more than enough.
And I would prefer dash, i.e. `wasm-shaper`

tschneidereit:
@asibahi this is very exciting! As a quick note, please don't hesitate to reach out to us in the Bytecode Alliance for any help we might be able to provide with Wasmtime. The Wasmtime maintainers are very responsive on [our Zulip](https://bytecodealliance.zulipchat.com/#narrow/stream/217126-wasmtime).

> `wasmtime` brings Cranelift and all its baggage around. `wasmi` is much lighter as a dependency and can be compiled to `no_std`.

Incidentally, we're working on both of those! There's an [RFC for adding an interpreter](https://github.com/bytecodealliance/rfcs/pull/35), and recently parts of the runtime gained [support for `no_std`](https://docs.wasmtime.dev/stability-platform-support.html#support-for-no_std). The latter doesn't yet include Cranelift, but that can be [fixed with some elbow grease](https://github.com/bytecodealliance/wasmtime/issues/1158#issuecomment-2150020984).

We also have minimal embeddings getting the size of Wasmtime down to about 2MB right now, with a clear path towards the hundreds of kilobytes range if needed.

Having said all this, `wasmi` is an excellent interpreter, and I'm sure @Robbepop would be excited to help with integrating it, too!

Robbepop:
Hi, I have just been made aware of this topic by @tschneidereit 's mention of me and haven't read everything (no time to do so rn). I am happy to help. Either solution, Wasmtime or Wasmi, should work fine since both mostly share the same API surface. It could even be possible to use both solutions for their different strengths.

src/hb/shape_wasm.rs
asibahi: Done. But I am not sure if that's better for the Ruqaa test, as it now doesn't distinguish between the wasm code failing and faulty output.

src/hb/buffer.rs
```diff
@@ -1556,6 +1557,7 @@ pub const HB_BUFFER_SCRATCH_FLAG_SHAPER0: u32 = 0x01000000;
 // pub const HB_BUFFER_SCRATCH_FLAG_SHAPER3: u32 = 0x08000000;
 
 /// A buffer that contains an input string ready for shaping.
+#[cfg_attr(feature = "wasm-shaper", derive(Clone))]
```
RazrFalcon, line 1560: Why do you have to clone the buffer? This is a very expensive operation. Better avoid it.

Cargo.toml
```diff
@@ -37,6 +38,9 @@ features = [
 default = ["std"]
 std = ["ttf-parser/std"]
 
+# experimental features live here.
```
RazrFalcon, line 41: No need for a comment and new line. Just put `wasm-shaper` right after `std`.

src/hb/shape.rs
```diff
@@ -20,7 +20,16 @@ pub fn shape(face: &hb_font_t, features: &[Feature], mut buffer: UnicodeBuffer)
         buffer.0.language.as_ref(),
         features,
     );
-    shape_with_plan(face, &plan, buffer)
+    #[cfg(not(feature = "wasm-shaper"))]
```
RazrFalcon, line 23: Shouldn't we check for the `Wasm` table here? I.e. this should be a runtime decision, not compile-time one.

If we have `Wasm` table, try `shape_with_wasm`, otherwise `shape_with_plan`.

src/hb/shape.rs
```diff
@@ -20,7 +20,16 @@ pub fn shape(face: &hb_font_t, features: &[Feature], mut buffer: UnicodeBuffer)
...
+    {
+        shape_with_plan(face, &plan, buffer)
+    }
+    #[cfg(feature = "wasm-shaper")]
+    {
+        // is there a way around the clone here?
```
RazrFalcon, line 29: Yes, `shape_internal` requires a mutable reference.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+    // We don't use font token and buffer token
+    _font: u32,
+    _buffer: u32,
+    // harfbuzz-wasm doesn't use the features pointer
+    _features: u32,
+    _num_features: u32,
+    // we dont have custom shapers (yet?).
+    _shaper: u32,
+) -> i32 {
+    // potentially we could read the shaper pointed to by `shaper`
```
RazrFalcon, line 530: Let's add the `log` dependency and warn when non `ot` shaper is requested.
Not sure how to report an error from this method. I guess simply log and skip the shape call.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    if memory.data(&caller).get(s as usize).is_none() {
+        return;
+    }
+    let bytes = &memory.data(&caller)[s as usize..];
+    let msg = CStr::from_bytes_until_nul(bytes)
+        .unwrap_or_default()
+        .to_string_lossy();
+
+    std::eprintln!("{msg}");
```
RazrFalcon, line 514: Let's add the `log` dependency.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+
+    mem_data[eom..pos_loc].copy_from_slice(bytemuck::cast_slice(&rb_buffer.info));
+    mem_data[pos_loc..end_loc].copy_from_slice(bytemuck::cast_slice(&rb_buffer.pos));
+
+    let buffer_contents = CBufferContents {
+        length: length as u32,
+        info: eom as u32,
+        position: pos_loc as u32,
+    };
+    let Ok(()) = memory.write(
```
RazrFalcon, line 434: I think we can replace it with `if memory.write().isErr() { return 0 }`

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+    y: f32,
+    pointtype: PointType,
+}
+impl OutlinePoint {
+    fn new(x: f32, y: f32, pointtype: PointType) -> Self {
+        OutlinePoint { x, y, pointtype }
+    }
+}
+
+unsafe impl bytemuck::Zeroable for OutlinePoint {}
```
RazrFalcon, line 237: Are you sure that the memory layout is the same as harfbuzz expects?

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+    QuadraticTo,
+    CubicTo,
+}
+
+#[repr(C)]
+#[derive(Clone, Copy, Debug)]
+struct OutlinePoint {
+    x: f32,
+    y: f32,
+    pointtype: PointType,
```
RazrFalcon, line 229: `pointtype` can be replaced with `kind`.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,548 @@
...
+    // len is the assigned heap memory. We should not allocate more than that.
+    // Should not assume we are not writing over anything.
+    let memory = caller.get_export("memory").unwrap().into_memory().unwrap();
+
+    let mut name = caller
+        .data()
+        .font
+        .glyph_name(GlyphId(glyph as u16))
+        .map(ToOwned::to_owned)
+        .unwrap_or(format!("g{glyph:4}"));
```
RazrFalcon, line 195: Does this produce `g0000`? Can you add a comment explaining this?

src/hb/shape.rs
asibahi: This is what this code does. The `Wasm` table is the first thing `shape_with_wasm` checks for, and if it is not there it returns `None`. The compile time check here is for the featuree as `shape_with_wasm` does not exist then.

src/hb/shape_wasm.rs
asibahi: I copied it as is from `harfbuzz-wasm`.

src/hb/shape_wasm.rs
asibahi: I dont understand what the problem is in returning that? The names in the test fonts are used to search for things like `.fina` (for Arabic) to find glyphs with specific properties. Any glyph with name `gXXXX` would be equally useless.

asibahi:
Fixed the bug and now the Ruqaa Font tests works if we do not compare clusters.

That'll teach me to read the standard library's documentation.

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,566 @@
...
+    buffer: &mut hb_buffer_t,
+) -> Option<()> {
+    // If font has no Wasm blob just return None to carry on as usual.
+    let wasm_blob = font
+        .raw_face()
+        .table(ttf_parser::Tag::from_bytes(b"Wasm"))?;
+
+    let data = ShapingData { font, plan, buffer };
+
+    let mut store = Store::new(&Engine::default(), data);
```
Robbepop, line 32: If the Wasm binaries are rather big and only parts of them are usually used (not all functions defined in them), then it might make sense to use [Wasmi's lazy compilation feature](https://docs.rs/wasmi/0.34.0/wasmi/struct.Config.html#method.compilation_mode) in order to improve startup times significantly. Expected speed-up for larger Wasm blobs is ~10-40x.

You can do this as follows:

```rust
let mut config = wasmi::Config::default();
engine.compilation_mode(wasmi::CompilationMode::Lazy);
let engine = wasmi::Engine::new(&config);

// now you can use engine
```

src/hb/shape_wasm.rs
```diff
@@ -0,0 +1,566 @@
...
+        .func_wrap("env", "face_copy_table", face_copy_table)
+        .ok()?
+        .func_wrap("env", "buffer_copy_contents", buffer_copy_contents)
+        .ok()?
+        .func_wrap("env", "buffer_set_contents", buffer_set_contents)
+        .ok()?
+        .func_wrap("env", "debugprint", debugprint)
+        .ok()?
+        .func_wrap("env", "shape_with", shape_with)
+        .ok()?;
```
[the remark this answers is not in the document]
asibahi: Where would be a good place to stash a cached `LinkerBuilder`, @RazrFalcon ? A `pub(crate) field in `hb_ot_shape_plan_t` maybe?

src/hb/shape_wasm.rs
asibahi: Thanks for the tip.

asibahi:
One thing I learned about recently:

Is `scale` from harfbuzz (the one expcted by this API) related to `pixels_per_em`? the `skrifa` crate seems to imply as much. If that's so we can perhaps just return the `pixels_per_em` if set, and if not, return UPEM?

src/hb/shape_wasm.rs
RazrFalcon: How so? Can you elaborate?

src/hb/shape.rs
RazrFalcon: But if there is no `Wasm` table you exit the shaping immediately, instead of calling the regular shaper. No?
I would expect something like:

```python
if has_wasm_table and run_wasm_shaper():
  return

run_regular_shaper()
```

src/hb/shape_wasm.rs
RazrFalcon: Ok then.

src/hb/shape_wasm.rs
RazrFalcon: Are you answering to the right comment? All I was asking for is to document `format!("g{glyph:4}")`, because it's not immediately obvious what it does. Maybe write it as  `format!("g{:4}", glyph)`

RazrFalcon:
>Is scale from harfbuzz (the one expcted by this API) related to pixels_per_em? 

Once again, there is no scale in rustybuzz. It was removed during port. It should always be 1.

>Also re: shape_with , doesn’t rustybuzz technically use the AAT shaper as well as OT?

In harfbuzz terminology this is still the `ot` shaper. See https://github.com/harfbuzz/harfbuzz/blob/main/src/hb-shaper-list.hh

behdad:
> > Is scale from harfbuzz (the one expcted by this API) related to pixels_per_em?
> 
> Once again, there is no scale in rustybuzz. It was removed during porting. It should always be 1.

It should be always units_per_em. And it's NOT related to pixels_per_em.

src/hb/shape_wasm.rs
asibahi: Oh I thought you meant specifically `g0000`. I will double check.

src/hb/shape.rs
asibahi: yeah `shape_with_wasm` fails if there is no `"Wasm"` table. The fallback to regular shaper [happens in `shape_with_plan` here](https://github.com/asibahi/rustybuzz/blob/84b9f58d97d19a947064a0b0cc995c6ca7560bca/src/hb/shape.rs#L60). You can see the regular tests pass just fine with the feature turned on.

src/hb/shape_wasm.rs
asibahi: It is not a very well formed thought, but the wasm shaper can fail in two ways (categories of ways, rather). Either `shape_with_wasm` returns with bad output, or it returns `None`, because of some wasm failure.  Since the tests in the tests folder only test the public API instead of private functions, they *always* return, as they fall back to the regular shaper. So one might not be able to tell if the wasm module crashed or not.

asibahi:
So I think this is done.

The full defined API in the [design document](https://github.com/harfbuzz/harfbuzz/blob/main/docs/wasm-shaper.md) is not implemented, but only the ones used by the [examples](https://github.com/harfbuzz/harfbuzz-wasm-examples).

RazrFalcon:
Thanks! Will be released soon. Will see how it will go.

PR #123, opened 2024-07-03
Add support for custom tests

LaurenzV:
Change the `gen-shaping-tests` script so that we can add our own, rustybuzz-specific tests. A first necessary steps toward fuzz-testing correctness against harfbuzz. Still in-progress.


LaurenzV:
It looks like a lot changed, but not really, just some files that had to be regenerated. The reason I split off MacOS tests is that otherwise, someone who doesn't have a MacOS system cannot regenerate the tests without removing the MacOS ones. And also, since we store `macos.tests` in rustybuzz anyway instead of using the `harfbuzz` one, it makes sense to treat them as custom tests as well.

Next step after this PR is to add some test for glyph buffer flags, where I've noticed some discrepancies with harfbuzz.

Last step will be to do fuzzing, I already have some ideas which I will expand on in a different PR.

RazrFalcon:
>And the Python script still feels pretty hacky, but it does its job...

Can you elaborate? All it does is converts HB tests to Rust. Way better then what HB doing by having a custom tests implementation.

LaurenzV:
> Can you elaborate? All it does is converts HB tests to Rust. Way better then what HB doing by having a custom tests implementation.

What I mean is the Python script itself is pretty ugly because of all the special cases that are hardcoded. :p

RazrFalcon:
Well, it has one job and does it well for 5 years. I do not see how we can have less hardcoded stuff and how it would help us to begin with.

PR #125, opened 2024-07-04
Fix some glyph flags bug

LaurenzV:
and use the new test infrastructure to test them!

Unfortunately, there is one bug remaining that I haven't been able to figure out. In rustybuzz, we have these lines of code inside of `state_machine_kerning`: https://github.com/RazrFalcon/rustybuzz/blob/e2f88315124d8a83b075979731510741c99feb4c/src/hb/kerning.rs#L186-L201

From what I can tell, this doesn't exist in `harfbuzz`. After debugging it seems like even for kerning, harfbuzz also uses the `drive` method which we use for morx, and it has a different mechanism for determining the flags: 

https://github.com/RazrFalcon/rustybuzz/blob/e2f88315124d8a83b075979731510741c99feb4c/src/hb/aat_layout_morx_table.rs#L223-L296

Unfortunately, I wasn't able to quite port that over because it has some fields which we don't seem to have in `state_machine_kern`. But I've still added a test and disabled it for now.

But I was able to squash two other bugs!


RazrFalcon:
>From what I can tell, this doesn't exist in harfbuzz.

Well, I ported it from HB, so it was existed at some time 😄 
Here it is: https://github.com/harfbuzz/harfbuzz/blob/05ef75c55340400d4b318bd24d742653bbf825d9/src/hb-aat-layout-common.hh#L754
I guess it was updated later.

LaurenzV:
Ah, thanks! I wasn't able to find it in previous commits, hopefully this will help. I'll try to give it another look.

RazrFalcon:
And this [change](https://github.com/harfbuzz/harfbuzz/commit/d8ea552d1095b92f44657ca8a90e4fc9fcdbd37a) is currently in `morx` and not in `common`, like in HB, for some reason. But it is still generic. Aka it's not morx specific. So we probably should move it to common.

RazrFalcon:
Here it is in RB: https://github.com/RazrFalcon/rustybuzz/blob/e2f88315124d8a83b075979731510741c99feb4c/src/hb/aat_layout_morx_table.rs#L160

I remember this code well, because I was arguing with behdad about it...

RazrFalcon:
Found it: https://github.com/harfbuzz/harfbuzz/commit/978918c32a66af41df86386510cf73a4c0e8d393

LaurenzV:
I'm super confused now. What is the difference between `apply_state_machine_kerning` in `aat_layout_kerx_table.rs` and `apply_state_machine_kerning` in `kerning.rs`?

RazrFalcon:
Good question. I guess I weren't able to make it generic enough and simply duplicated the implementation.
You can try merging it into one function.

RazrFalcon:
Sure.

Issue #132, opened 2024-09-11
RTL shaping output not correct.


LaurenzV:
Using Noto Sans CJK, the following input with `--direction rtl` yields the same output as in harfbuzz:
```
shape /Library/Fonts/NotoSansCJKsc-Regular.otf --direction rtl "你好"
gid14225=1+1000|gid9987=0+1000
hb-shape /Library/Fonts/NotoSansCJKsc-Regular.otf --direction rtl "你好"
gid14225=1+1000|gid9987=0+1000
```

However, once I slightly adapt the input string the output suddenly is wrong and not reversed anymore:
```
shape /Library/Fonts/NotoSansCJKsc-Regular.otf --direction rtl "你好，"
gid9987=0+1000|gid14225=1+1000|gid59058=2+1000
hb-shape /Library/Fonts/NotoSansCJKsc-Regular.otf --direction rtl "你好，"
gid59058=2+1000|gid14225=1+1000|gid9987=0+1000
```

PR #133, opened 2024-09-11
Fix issue in `reverse_groups` method

LaurenzV:
Turns out I just copied it wrongly nested from harfbuzz... Whoops. 😅 

Here is the original method, for reference:
```cpp
  template <typename FuncType>
  void reverse_groups (const FuncType& group,
		       bool merge_clusters = false)
  {
    if (unlikely (!len))
      return;

    unsigned start = 0;
    unsigned i;
    for (i = 1; i < len; i++)
    {
      if (!group (info[i - 1], info[i]))
      {
	if (merge_clusters)
	  this->merge_clusters (start, i);
	reverse_range (start, i);
	start = i;
      }
    }
    if (merge_clusters)
      this->merge_clusters (start, i);
    reverse_range (start, i);

    reverse ();
  }
```


RazrFalcon:
>Turns out I just copied it wrongly nested from harfbuzz...

Yeah, if only HB used spaces for indents... I had similar issues as well. Eventually I simply reformatted the whole project during initial porting.

Is this related to #132?

LaurenzV:
> Is this related to https://github.com/RazrFalcon/rustybuzz/issues/132?

Yes, it fixes that exact issue.

RazrFalcon:
Thanks!

Issue #136, opened 2024-10-03
hb_font_t::set_variations sets wrong coordinate values for fonts with avar tables and multiple axes


maxmelander:
When calling `hb_font_t::set_variations`, `set_variation` in ttf-parser will be called once for each axis. 
If the font uses an avar table, `avar.map_coordinates` will be called for all axes each time `set_variation` is called. This causes the avar table to be applied multiple times to the same coordinate.

Repro:
Font: [Recursive](https://fonts.google.com/specimen/Recursive)
Variations: 
[('slnt', 0.0), ('wght', 400.0), ('CASL', 0.0), ('CRSV', 0.0), ('MONO', 0.0)]

Setting the variations in this order will cause the normalized coordinate value for 'wght' to go something like this:
```
0. Sets 'slnt'
1. Sets 'wght' to NormalizedCoordinate(2340)
1.1 AVAR table is applied and 'wght' is set to NormalizedCoordinate(4117)

2. Sets 'CASL' to NormalizedCoordinate(0)
2.1. AVAR table is applied to everything including `wght` again. `wght` is now set to NormalizedCoordinate(5221)

...
```
After the whole loop `wght` ends up with a value of NormalizedCoordinate(6593)

You can also repro this on fonts with a single variable axis like [Spline Sans](https://fonts.google.com/specimen/Spline+Sans?query=spline+sans) by calling set_variations with a bunch of variations that do not exist in the font. 

Hope this makes sense:)

RazrFalcon:
Does `harfbuzz` produce the correct output?

RazrFalcon:
This change breaks `in_house::variations_005` test. So it either `harfbuzz` is wrong or there is something else going on.

LaurenzV:
I get the same result when using `coretext` as the shaper, so I don't think it's a harfbuzz issue.

RazrFalcon:
@LaurenzV but we implement only `ot` shaper, so it should not break, no?

LaurenzV:
Not quite sure what you mean, but when I try the suggested `ttf-parser` patch I also get a failure for `in_house::variations_005`, like you, so there seemingly is something wrong with that patch.

LaurenzV:
Oh, I mean shaping with harfbuzz but with `--shaper=coretext`

maxmelander:
Looking at this commit, which adds the failing test, I think the error in my patch is that I don't handle multiple axes with the same tag:
[...]

PR #137, opened 2024-10-04
Add test covering incorrect application of avar table

maxmelander:
Adds test to cover the issue where the avar table was beeing applied to the same coordinate multiple times when calling set_variations ( #136 ). 

This test will pass when RazrFalcon/ttf-parser@1cac7c71c179660b927b0add4b061b9fca3d6d9d is taken into use.


RazrFalcon:
I have updated `ttf-parser`, so the test should not fail anymore.
But since we auto-generate tests, there are a couple more steps needed.

1. Put your test data into `tests/custom/bugs.tests`.
2. Move the font to `tests/fonts/rb_custom`.
3. Move `variations_006` to `tests/shaping/custom.rs` as `bugs_002`.

That should be enough.

RazrFalcon:
Thanks!

PR #140, opened 2024-10-24
Clippy fixes

asibahi:
The first commit contains a bulk of basic lints (avoiding double refs, needless returns, needless `format!`, and so on).

The other commits are divided by lint they apply.


RazrFalcon:
`collapsible if` should probably be ignored. We try to match harfbuzz's code style when possible. So the code should look as close to harfbuzz as possible, at least in theory.

alerque:
We can configure the lint to be ignored project wide if it is interfering with keeping the cognitive load low when comparing code bases.

asibahi (commented):
These are the remaining lints. I wonder where a good place to add them would be:

```rust
// Quiet Clippy lints left to match hb style
#![allow(clippy::collapsible_if)]
#![allow(clippy::collapsible_else_if)]
#![allow(clippy::comparison_chain)]
#![allow(clippy::needless_range_loop)]
#![allow(clippy::non_canonical_partial_ord_impl)]
#![allow(clippy::doc_lazy_continuation)] // <-- this one is about docs, not code style
#![allow(clippy::upper_case_acronyms)]
#![allow(clippy::too_many_arguments)]
#![allow(clippy::wildcard_in_or_patterns)]
#![allow(clippy::identity_op)]
#![allow(clippy::mut_range_bound)]
#![allow(clippy::enum_variant_names)]
#![allow(clippy::manual_range_patterns)]
#![allow(clippy::type_complexity)]
#![allow(clippy::wrong_self_convention)]
#![allow(clippy::match_like_matches_macro)]
```

RazrFalcon:
Looks fine to me, but let's also disallow `(a..b).contains(&c)`. Since it's not available in C++. Should make porting easier.

RazrFalcon:
Looks good. Will wait for someone to merged it.

src/hb/ot_shaper_vowel_constraints.rs
```diff
@@ -365,14 +365,13 @@ pub fn preprocess_text_vowel_constraints(buffer: &mut hb_buffer_t) {
             while buffer.idx + 1 < buffer.len {
                 #[allow(unused_mut)]
                 let mut matched = false;
-                match buffer.cur(0).glyph_id {
-                    0x112B0 => match buffer.cur(1).glyph_id {
+                if buffer.cur(0).glyph_id == 0x112B0 { 
+                    match buffer.cur(1).glyph_id {
```
khaledhosny, line 369: This is a generated file, so this will be overwritten the next time this file is re-generated.

asibahi: I guess the python script could add the lint somewhere at the top of the file? 

```py
print('#![allow(clippy::single_match)]')
```

PR #144, opened 2024-11-09
Fix overflow when adding delta

LaurenzV:
Same as #143, but with a minimal test case.


alerque:
I'm new to helping out with this project and the general flow of things, but as a *general* rule of thumb from other projects I find it helpful to have fixes and tests in separate commits. Would you mind if I merge Simon's fix first then this on top of it as test commit? I can handle the rebasing, just asking if that general handling of code changes vs. testing is alright by you.

khaledhosny:
In HarfBuzz we pretty much always have tests added in the same commit that fixes them/implments the feature they test.

LaurenzV:
Both are fine to me, your choice! :)

alerque:
> In HarfBuzz we pretty much always have tests added in the same commit that fixes them/implments the feature they test.

Sure, many projects do this and there isn't a hard and fast rule about the right way to do it.

Personally I like the workflow where one can easily `git revert <commit id of fix commit>` and quickly prove that the test case actually catches the bug it was intended to. With the fix and the test in the same commit one has to manually work out from reviewing the patch something to twiddle to verify the test works. Probably not relevant to *this* issue, but having the tests available to cherry pick also makes it easier to play around with potential alternative fixes.

alerque (approved):

PR #148, opened 2024-11-12
Propose publishing a patch release

alerque:
Since we have a bugfix for an actual crash caused by a type mismatch from the HarfBuzz version this is currently synced with, I am proposing a patch level release now. Syncing with HB 10.1.0 can happen whenever somebody gets around to doing the porting, but I don't see anything else upcoming that we need to wait for to ship this patch. Any objections to my going ahead and publishing this?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "rustybuzz"
-version = "0.20.0"
+version = "0.20.1"
 authors = [
     "Caleb Maclennan <caleb@alerque.com>",
     "Laurenz Stampfl <laurenz.stampfl@gmail.com>",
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -1,7 +1,12 @@
 [package]
 name = "rustybuzz"
-version = "0.20.0"
-authors = ["Yevhenii Reizner <razrfalcon@gmail.com>"]
+version = "0.20.1"
+authors = [
+    "Caleb Maclennan <caleb@alerque.com>",
+    "Laurenz Stampfl <laurmaedje@gmail.com>",
```
LaurenzV, line 6: Email is from a different Laurenz 👀 Mine is laurenz.stampfl@gmail.com

alerque: Yikes. Also not the first time I've conflated you!

LaurenzV (approved):

PR #149, opened 2024-11-17
bench: switch to criterion

jcdickinson:
This is a "PR that could have been an issue," I'm opening this PR to determine if this is a direction that the project wants to take.

Switches the benchmark framework to Criterion. The default Rust std test framework has a hardcoded goal execution time of 1ms for each benchmark. This makes it unsuitable for sample-based profiling because the benchmark doesn't run for very long.

Adds a utility script to run tests under a profiler.

### DIFF
--- a/benches/.gitignore
+++ b/benches/.gitignore
@@ -1,0 +1,4 @@
+target
+perf.data
+perf.data.old
+Cargo.lock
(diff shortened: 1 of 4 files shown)


RazrFalcon:
I'm aware of `criterion`, but it's an overkill for our use case. And it's a very heavy dependency. The current implementation is more than enough.

jcdickinson:
> The current implementation is more than enough.

@RazrFalcon what about the profiling issue I mentioned in the description? 1ms of sampling is certainly completely inadequate to create a high quality profile: https://github.com/rust-lang/rust/blob/master/library/test/src/bench.rs#L132

PR #153, opened 2024-11-29
Relax dependency pinning from patch to minor level only

alerque:
One of the things I've found frustrating trying to update dependencies in my own projects (and 3rd party ones) is keeping the dependency tree free of duplicate copies of the same crate at different pinned versions. This is a notable problem for this crate in particular because it has been a number of deepnesses pinned to exact patch versions and is frequently found in conjunction with other crates that do the same. An example of this is Typst (which uses rustybuzz directly) where it depends on resvg, which in turn depends on RustyBuzz and ttf-parser as well.

I propose relaxing the main manifest to only define dependencies at the minor version level. This will make downstream dependency trees quite a bit easier to manage and allow more frequent version bumps for some projects without bloating the build tree.

In the mean time we still have a tracked lock manifest with exact patch versions for what we have most recently tested and released.

Any feelings on this guys?


LaurenzV:
I thought that even if you specify the minor verison in `Cargo.toml`, as long as you don't specify the dependency with `=` the minor version won't be pinned? But if that's not the case, I'm fine with this change.

LaurenzV (approved):

alerque:
> as long as you don't specify the dependency with `=` the minor version won't be pinned

True, but the default [caret requirements](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#caret-requirements) syntax allows semver compatible updates. If you specify `1.2.3` it is allowed to update to any `1.2.x` release. If you specify `1.2` then it is allowed to update to any `1.x` release. Relaxing our dependencies from patch level updates to minor version level is what this PR is about. This should be fine as long as no dependencies start outputting semver-breaking changes in non-breaking releases. In that case all bets are kind of off anyway, and it could just as easily happen in a patch release too.

PR #152, opened 2024-11-29
Bump ttf-parser

alerque:
...and keep such bumps from updating the lockfile manifest format and hence prematurely bumping our MSRV.

PR #154, opened 2024-11-29
Bump wasmi crate dependency

alerque:
The wasmi dependency is only used for the non-default WASM shaper featuer. I built and ran tests with this so it is probably fine, but @asibahi I thought I'd run this by you to make sure there isn't some reason to hold this back.

This will need a lock file bump to go with it, but I'll sort that out on merge dependency on #152 and #153.


asibahi:
I have no problem with this.

Dependncy update aside, the module could stand to be made more efficient, mostly by caching the VM instance to avoid creating a new one for every call to `shape`. But I think unless this has more users (and user feedback) it is fine as is.

The API could definitely be improved, however, but this is something from harfbuzz's PoV, really.

LaurenzV (approved):

PR #157, opened 2025-03-06
Add tests for COLRv1 glyph extents

valadaptive:
About half of them currently fail.


LaurenzV:
Thanks! I was aware that there were some failing tests as I tried the same before, but I didn't manage to look into what caused them.

tests/custom/colr.tests
```diff
@@ -0,0 +1,20 @@
+tests/fonts/rb_custom/test_glyphs-glyf_colr_1_no_cliplist.ttf;--show-extents;U+F0100,U+F0101,U+F0102,U+F0103;[linear_repeat_0_1=0+1000<100,950,800,-700>|linear_repeat_0.2_0.8=1+1000<100,950,800,-700>|linear_repeat_0_1.5=2+1000<100,950,800,-700>|linear_repeat_0.5_1.5=3+1000<100,950,800,-700>]
```
LaurenzV, line 1: Could you remove the expected glyphs from each line? Since they are generated manually by running them against harfbuzz, they shouldn't be necessary for generating the shaping tests, IIRC.

LaurenzV (changes requested):

tests/custom/colr.tests
valadaptive: I've added `--no-glyph-names` to the test arguments.

LaurenzV: Ah no, I meant completely removing the last part from each line. For example, the first line should be changed from
`tests/fonts/rb_custom/test_glyphs-glyf_colr_1_no_cliplist.ttf;--show-extents --no-glyph-names;U+F0100,U+F0101,U+F0102,U+F0103;[8=0+1000<100,950,800,-700>|9=1+1000<100,950,800,-700>|10=2+1000<100,950,800,-700>|11=3+1000<100,950,800,-700>]`
to
`tests/fonts/rb_custom/test_glyphs-glyf_colr_1_no_cliplist.ttf;--show-extents --no-glyph-names;U+F0100,U+F0101,U+F0102,U+F0103;` (with trailing semicolon)

The expected glyphs are generated by invoking hb-shape from the CLI, so there is no need to add them to `colr.tests`, see for example also `bugs.tests` or `fuzzer.tests`.

valadaptive: Ah, I see. I've done that now. I've removed `--no-glyph-names` from the arguments as well; it doesn't seem to be in many of the other tests and the glyph names are descriptive and could be useful for debugging why the tests are failing.

LaurenzV:
Thanks!

PR #158, opened 2025-03-06
Fix COLR glyph extents calculations

valadaptive:
I added custom tests for COLR glyph extents using the font mentioned in the TODO (see #157). Many of the tests failed, so I fixed the bugs causing them to do so:
- The `union_` and `intersect` methods were copied incorrectly from HarfBuzz and overwrote the destination bounds entirely.
- We previously bailed out early if the COLR table was version 0, but we don't actually need to do this.
- COLRv0 glyphs in general (which can seemingly appear even if the version of the *table* is 1?) currently don't work because of an issue in ttf-parser (https://github.com/harfbuzz/ttf-parser/pull/185). I've included tests for them anyway, which can be enabled if/when that PR is merged. I've confirmed that they pass when using a local ttf-parser with that change included.
- We previously bailed out entirely if we couldn't get the extents for a COLR glyph from the COLR table, but HarfBuzz falls back to looking at other tables.
- HarfBuzz intersects clip bounds with existing ones when `push_clip` is called. This doesn't seem to affect any tests, but I ran into it while tracking down the harder-to-spot first issue (`union_` and `intersect`).


LaurenzV:
That seems good, thanks!

PR #159, opened 2025-03-07
Make WASM shaper's `OutlinePoint` type `NoUninit` instead of `Pod`

valadaptive:
The `Pod` trait cannot be implemented on enums, since they don't allow any bit pattern. `NoUninit` can, though the enum needs to specify its integer representation explicitly. (See `font_copy_glyph_outline` in https://github.com/harfbuzz/harfbuzz/blob/fe8a397/docs/wasm-shaper.md)


LaurenzV:
@asibahi Does that seem reasonable to you? If I remember correctly you wrote that.

asibahi:
Looks fine to me, from what I understand in the bytemuck docs

